PR #134, opened 2025-05-23
Jcb/ix tests

johnchandlerburnham:
big refactor of the ix compiler, the ixon format and implementation of the decompiler

roundtrip compilation/decompilation of Std, Lean and Ix itself now working (modulo Expr.mdata, which is currently not captured by Ixon metadata, but arguably should be)


arthurpaulino (approved):
I'm not sure about the coding style that deviates from the common Lean 4 usage. But the sources look good. Huge achievement!

PR #136, opened 2025-05-28
feat: Blake3 compression in Lean

storojs72:
This is a Lean port of Blake3 compression algorithm from [following](https://github.com/BLAKE3-team/BLAKE3/blob/master/reference_impl/reference_impl.rs#L74) Rust reference. It is slightly adapted in order to be extended to a trace generation algorithm that we need for Archon modules.

### DIFF
--- a/Tests/Blake3.lean
+++ b/Tests/Blake3.lean
@@ -2,16 +2,16 @@ import LSpec
 
 namespace Blake3
 
-def IV : Array UInt32 := #[0x6A09E667, 0xBB67AE85, 0x3C6EF372, 0xA54FF53A, 0x510E527F, 0x9B05688C, 0x1F83D9AB, 0x5BE0CD19]
-def MSG_PERMUTATION : Array (Fin 16) := #[2, 6, 3, 10, 7, 0, 4, 13, 1, 11, 12, 5, 9, 14, 15, 8]
-def A : Array (Fin 32) := #[0, 1, 2, 3, 0, 1, 2, 3]
-def B : Array (Fin 32) := #[4, 5, 6, 7, 5, 6, 7, 4]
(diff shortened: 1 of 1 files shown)


Tests/Blake3.lean
```diff
@@ -0,0 +1,140 @@
...
+def round (state: Array UInt32) : Array UInt32 :=
+  let state := roundNoPermute state
+  permute state
+
+
+def compress (cv blockWords : Array UInt32) (counter : UInt64) (blockLen flags : UInt32) : Array UInt32 :=
+  let counterLow := UInt32.ofBitVec (counter.toBitVec.truncate 32)
+  let counterHigh := UInt32.ofBitVec ((counter.shiftRight 32).toBitVec.truncate 32)
+
+  let state := #[
```
arthurpaulino, line 70: For a less error-prone version, you can write something like
```lean
let state := (cv.extract 0 8) ++ (IV.extract 0 4) ++
  #[counterLow, ..., flags] ++ (blockWords.extract 0 16)
```

Tests/Blake3.lean
```diff
@@ -21,7 +21,7 @@ def permute (state: Array UInt32) : Array UInt32 :=
 
   left ++ permuted
 
-def transition (state: Array UInt32) (j : Fin 8) : Array UInt32 :=
+def transition (state: Vector UInt32 32) (j : Fin 8) : Vector UInt32 32 :=
   let aIn := state[(A[j])]!
   let bIn := state[(B[j])]!
   let cIn := state[(C[j])]!
```
arthurpaulino, line 27: I think you're able to remove these exclamation marks now

arthurpaulino (approved):
Great start!

PR #162, opened 2025-06-27
Aiur multiplication

gabriel-barrett:


Ix/Archon/TowerField.lean
```diff
@@ -38,6 +38,9 @@ opaque mulUInt64InBinaryField : @& UInt64 → @& UInt64 → UInt64
 @[extern "rs_pow_u64_in_binary_field"]
 opaque powUInt64InBinaryField : @& UInt64 → @& UInt64 → UInt64
 
+@[extern "c_rs_pow_u128_in_binary_field"]
+opaque powUInt128InBinaryField : @& UInt128 → @& UInt64 → UInt128
```
arthurpaulino, line 42: In cases like this it's good to be explicit about which argument is what. I can change it on my incoming PR.

arthurpaulino (approved):
Thanks!!

PR #177, opened 2025-07-02
chore: Introduce Gadget.provide function

storojs72:
Since the way of providing data in gadget implementations is not going to change in the meanwhile, let's have it as a separate function to hide some of Aiur specifics from the user


arthurpaulino (approved):
Thanks!

PR #187, opened 2025-07-28
chore: Configure Iroh with `net` feature

samuelburnham:
Restores Iroh functionality under a `net` feature that can be configured in Lean via the `IX_NET` env var. When the `net` feature is disabled or we compile on MacOS aarch64, the unsupported error message is printed instead.

Also separates the dummy functions into a separate module to cut down on `cfg`s.

### DIFF
--- a/Ix/Iroh/Transfer.lean
+++ b/Ix/Iroh/Transfer.lean
@@ -1,7 +1,7 @@
 namespace Iroh.Transfer
 
 @[never_extract, extern "c_rs_iroh_send"]
-opaque sendBytes' : @& ByteArray → Except String Unit
+private opaque sendBytes' : @& ByteArray → Except String Unit
 
 def sendBytes (bytes: @&ByteArray) : IO Unit :=
   match sendBytes' bytes with


Cargo.toml
```diff
@@ -13,13 +13,13 @@ rayon = "1"
...
-# iroh = "0.34.0"
-# tokio = "1.44.1"
-# iroh-base = "0.34.0"
-# iroh-blobs = { version = "0.34.0", features = ["rpc"] }
+# Iroh dependencies
+bytes = { version = "1.10.1", optional = true }
+tokio = { version = "1.44.1", optional = true }
+iroh = { version = "0.34.0", optional = true }
+iroh-base = { version = "0.34.0", optional = true }
+iroh-blobs = { version = "0.34.0", features = ["rpc"], optional = true }
```
arthurpaulino, line 21: I've been trying to follow the convention:
* If there's a major release (v1.0.0 and beyond), I just specify that major version
* Otherwise, I specify the version up to the minor identifier, leaving patches on automatic

This way `cargo update` can pick relevant updates with very limited chances of breakage.

Does this make sense?

arthurpaulino (approved):
Thanks! The diff is already good in this state as it accomplishes its goal. You can apply minor patches as you see fit.

Question, is everything good on the Nix side too?

Cargo.toml
samuelburnham: For post-1.0 crates, which are considered stable, `cargo update` will pull in new minor and patch versions regardless of whether the minor or patch version is specified. Thus, I think it makes more sense to specify at least the minor version as the minimum version that is expected to compile. `Cargo.lock` will specify the exact version, but it is often subject to change or regeneration so it seems clearer to the reader what the intent is. For an especially stable/well-maintained crate like `serde` there's probably not much benefit to specifying a minor version though.

For pre-1.0 crates, which are considered unstable, `cargo update` will only pull in new patch versions, regardless of whether the patch version is specified. A new minor version pre-1.0 will almost always include breaking changes, so it doesn't make sense to update automatically.

In either case, specifying the patch version doesn't affect anything for `cargo update`.  So my opinion is that it usually doesn't matter due to the liberal behavior of `cargo update` but I'd err on the side of over-specifying version numbers for readability. It's also worth nothing `cargo add` specifies minor and patch version for both pre- and post-1.0 crates.

See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax

c/iroh.c
```diff
@@ -1,35 +1,36 @@
...
-//         free(buffer);
-//     }
-//     rs__c_result_unit_string_free(result);
+    lean_object *io_result;
+    if (result->is_ok) {
+        io_result = lean_io_result_mk_ok(buffer);
+    }
+    else {
+        io_result = lean_mk_io_user_error(lean_mk_string(result->data));
+        // TODO: This frees an invalid pointer, but if removed causes an uncaught exception
```
arthurpaulino, line 30:
```suggestion
```
This is not an invalid pointer, as `lean_alloc_sarray` truly allocates memory (that must be freed explicitly, as stated in the comment above).

Ix/Iroh/Transfer.lean
```diff
@@ -1,9 +1,19 @@
 namespace Iroh.Transfer
 
 @[never_extract, extern "c_rs_iroh_send"]
-opaque sendBytes : @& ByteArray → IO Unit
+opaque sendBytes' : @& ByteArray → Except String Unit
```
arthurpaulino, line 4: Last thing: make `sendBytes'` private

PR #191, opened 2025-07-31
feat: Add Aiur benchmarks

### LINKED ISSUE #171: Aiur benchmarks
Use the `Benchmarks` API to benchmark Aiur.

samuelburnham:
This PR builds off of #186 and runs end-to-end as well as individual benchmarks of each step in Aiur proof generation.

Also enables support for benching both pure and `IO` functions so that we can selectively apply the `blackBoxIO` trick to only pure functions when using `bench`. If the function to be benchmarked already returns an `IO` type then we can use use the `benchIO` function instead.

Closes #171

### DIFF
--- a/Benchmarks/Main.lean
+++ b/Benchmarks/Main.lean
@@ -57,30 +57,69 @@ def friParameters : Aiur.FriParameters := {
   proofOfWorkBits := 20
 }
 
-def fib (n : Nat) : Nat :=
-  match n with
-  | 0 => 0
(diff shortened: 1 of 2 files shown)


arthurpaulino:
`Benchmarks.Main` needs to be adjusted after rebasing on top of #188. We now have `CommitmentParameters` and `FriParameters`

arthurpaulino (approved):

PR #214, opened 2025-09-10
chore: Use Garnix cache in Nix dev shell tests

samuelburnham:
Drops the Nix devshell tests job from over 20 min to 3-4 min


arthurpaulino (approved):
Thanks!!

PR #217, opened 2025-09-17
chore: Update Lean to v4.23.0

samuelburnham:


arthurpaulino (commented):
This is great! I just have one ask: please drop the changes that just remove blank spaces to avoid conflicts with John's work.
Conflicts aren't the end of the world but it's good to avoid overhead when we can.

arthurpaulino (approved):
Thank you very much!

PR #234, opened 2025-10-20
chore: Use pre-built Lean binary in Nix build

samuelburnham:
This PR switches to the pre-built lean4-nix binary introduced by https://github.com/lenianiva/lean4-nix/pull/70, which reduces the Lean Nix package build time from >10 min to <30 sec.

Garnix is now only needed to cache the Ix package and its dependencies e.g. Batteries, which would take too long to rebuild on every Nix CI run.

Also formats the Nix files with https://github.com/kamadorueda/alejandra

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -107,7 +107,6 @@
 
         # Provide a unified dev shell with Lean + Rust
         devShells.default = pkgs.mkShell {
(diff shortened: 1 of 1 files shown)


arthurpaulino (approved):
Approving to unblock

PR #236, opened 2025-10-20
feat: Add Iroh byte storage server example

samuelburnham:
This PR adds an example Iroh server and client for pinning and retrieving bytes via the Ix CLI, with nontrivial C FFI glue to illustrate how to pass data between Lean and Rust.

- `ix serve` starts the server, which prints the subsequent commands for the client
- `ix connect put` sends bytes to the server and receives their Blake3 hash
- `ix connect get` retrieves bytes from the server based on their hash

### DIFF
--- a/src/lean/ffi/mod.rs
+++ b/src/lean/ffi/mod.rs
@@ -47,8 +47,15 @@ pub(super) fn drop_raw<T>(ptr: *mut T) {
     drop(t);
 }
 
+// Only used in the Iroh client for the moment
 #[inline]
-#[allow(dead_code)]
+#[cfg_attr(
+    any(
+        not(feature = "net"),
+        all(target_os = "macos", target_arch = "aarch64")
+    ),
+    allow(dead_code)
+)]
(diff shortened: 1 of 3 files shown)


src/lean/ffi/mod.rs
```diff
@@ -58,7 +49,7 @@ pub(super) fn drop_raw<T>(ptr: *mut T) {
 
 #[inline]
 #[allow(dead_code)]
-pub(super) fn raw_to_str<'a>(ptr: *const c_char) -> &'a str {
+pub(crate) fn raw_to_str<'a>(ptr: *const c_char) -> &'a str {
     let c_str = unsafe { CStr::from_ptr(ptr) };
     c_str.to_str().expect("Invalid UTF-8 string")
 }
```
arthurpaulino: The fact that you've changed `raw_to_str` from `pub(super)` to `pub(crate)` makes me think that this function is now being used. In that case, it's no longer `dead_code`.

arthurpaulino (approved):
The FFI code looks great. Thanks!!
I'm approving with minor suggestions.

arthurpaulino (approved):

PR #246, opened 2025-10-28
Implement Ixon datatypes and serialization in Rust

johnchandlerburnham:


src/ixon/meta.rs
```diff
@@ -1,8 +1,46 @@
 use crate::ixon::address::Address;
-use crate::ixon::name::Name;
+//use crate::ixon::name::Name;
 use crate::ixon::nat::Nat;
 //use crate::ixon::serialize::Serialize;
```
arthurpaulino: This line can be deleted too.

src/ixon/meta.rs
```diff
@@ -21,17 +59,18 @@ pub enum ReducibilityHints {
...
+    Muts(Vec<Vec<Address>>),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Metadata {
-    pub map: Vec<(Nat, Vec<Metadatum>)>,
+    pub nodes: Vec<Metadatum>,
 }
 
 //impl Serialize for BinderInfo {
```
arthurpaulino: `Serialize` is now implemented in `ixon.rs`, right? I think these commented lines can be deleted.

arthurpaulino (commented):
We should fully migrate to `crate::lean::nat`. Other than that, my comments are all about further cleanup.

johnchandlerburnham:
> We should fully migrate to `crate::lean::nat`. Other than that, my comments are all about further cleanup.

`meta.rs` was actually supposed to have been removed already, thanks for catching that

arthurpaulino (approved):

PR #265, opened 2025-11-14
ci: Add comparative benchmark report

samuelburnham:
- Adds a benchmark Markdown report that prints to the terminal and to a file, controlled by the `report := true` `Config` setting and the `BENCH_REPORT` env var.
- Adds support for one-shot benchmarks for more expensive runs that don't need statistical analysis. Controlled by the `oneShot := true` `Config` setting
- Adds a GitHub Actions workflow to run a comparative benchmark by commenting `!benchmark` on a pull request, similar to https://github.com/lurk-lab/lurk-beta/blob/main/.github/workflows/bench-pr-comment.yml. The usage is
```
!benchmark <bench-A> <bench-B> <...>
ENV_A=var
ENV_B=var
<...>
```
If just `!benchmark` is specified then it will just run `bench-aiur` with no additional env vars. Each benchmark will create a separate CI job and PR comment [example](https://github.com/argumentcomputer/ci-lab/pull/95#issuecomment-3535048281). In addition, if the bench target does not exist in the `lakefile` on `main`, it will just run the benchmark on the PR and not compare [example](https://github.com/argumentcomputer/ci-lab/pull/95#issuecomment-3535063956).

Successful runs:
https://github.com/argumentcomputer/ci-lab/pull/95#issuecomment-3533854953


arthurpaulino (approved):
Nice! This is the kind of feature that we have to test in order to provide more accurate feedback.

PR #267, opened 2025-11-17
chore: Clean up Benchmark API

samuelburnham:
- Refactors the Blake3 benchmark to use the Benchmark API with `oneShot := true`, which enables printed reports and integration with the PR comment action
- Refactors the Benchmark API to be more idiomatic and deduplicate code, mainly merging one-shot and sampled benchmarks
- Moves benchmark files under `.lake/benches/<group>/<name>/<base|change|new>`, whereas before the group name wasn't included. This helps organize related benchmarks and aligns with Criterion.rs
- Fixes PR comment action to get the correct PR commit hash

Once merged we can test `!benchmark bench-aiur bench-blake3`, though the throughput measurement of the latter isn't supported within the PR comment yet.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -69,11 +69,11 @@ jobs:
       - uses: actions/checkout@v5
         # Get base branch of the PR
       - uses: xt0rted/pull-request-comment-branch@v2
-        id: base-branch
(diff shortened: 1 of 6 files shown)


samuelburnham:
!benchmark

PR #300, opened 2026-02-06
Add kernel typechecker

johnchandlerburnham:
~Implement a Lean 4 kernel typechecker using a DAG representation with BUBS (Bottom-Up Beta Substitution) for efficient reduction. The kernel operates on a mutable DAG rather than tree-based expressions, enabling in-place substitution and shared subterm reduction.~


Ix/Kernel/Level.lean
```diff
@@ -0,0 +1,131 @@
...
+/-- Reduce a level to normal form. -/
+def reduce : Level m → Level m
+  | .succ u => .succ (reduce u)
+  | .max a b => reduceMax (reduce a) (reduce b)
+  | .imax a b =>
+    let b' := reduce b
+    match b' with
+    | .zero => .zero
+    | .succ _ => reduceMax (reduce a) b'
+    | _ => .imax (reduce a) b'
```
gabriel-barrett: Shouldn't this be `| _ => reduceIMax (reduce a) b'`

johnchandlerburnham: yes, that's a bug

PR #308, opened 2026-02-18
ci: Build all Lake targets

samuelburnham:
Allows new benchmarks and targets to be automatically tested by CI, similar to `cargo build --all-targets`

Follow-up work: Update lean4-nix to allow packaging Lake scripts, so we can run this in Nix CI

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -1,7 +1,5 @@
 import Lake
-
 open System Lake DSL
-
 
 package ix where
   version := v!"0.1.0"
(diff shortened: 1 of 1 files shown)


lakefile.lean
```diff
@@ -58,8 +57,6 @@ lean_exe «bench-shardmap» where
 
 end Benchmarks
 
-lean_lib Apps
```
arthurpaulino: Won't this prevent Files in `Apps` from importing each other?

samuelburnham: It didn't build because there was no `Apps.lean` file so I added an empty one

PR #319, opened 2026-03-04
feat: Implement `u32_less_than` comparison primitive for Aiur

arthurpaulino:
Find witness c such that a + c + 1 = b + carry · 2^32, then a < b ⟺ carry = 0. Range-check all 12 byte auxiliaries via a new Bytes2 range-check channel that validates two field elements are bytes with no output columns.

### DIFF
--- a/src/aiur/constraints.rs
+++ b/src/aiur/constraints.rs
@@ -2,10 +2,10 @@ use multi_stark::{
   builder::symbolic::{SymbolicExpression, var},
   lookup::Lookup,
   p3_air::{Air, AirBuilder, BaseAir},
-  p3_field::PrimeCharacteristicRing,
+  p3_field::{Field, PrimeCharacteristicRing},
   p3_matrix::Matrix,
 };
-use std::ops::Range;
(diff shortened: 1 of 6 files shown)


src/aiur/constraints.rs
```diff
@@ -448,6 +449,87 @@ impl Op {
...
+        // Carry chain: a + c + 1 = b + carry * 2^32
+        let base_inv = G::from_u64(256).inverse();
+        let mut carry = Expr::ONE; // initial carry = 1 for strict less-than
+        for k in 0..4 {
+          let sum = x_bytes[k].clone() + y_bytes[k].clone() + carry;
+          carry = (sum - z_bytes[k].clone()) * base_inv;
+          state
+            .constraints
+            .zeros
+            .push(sel.clone() * (carry.clone() * (carry.clone() - Expr::ONE)));
```
gabriel-barrett: nit: although this constraint is correct, the standard boolean constraint is `b * (1 - b)`, not `b * (b - 1)`

PR #337, opened 2026-03-19
Type alias

gabriel-barrett:
Adds type aliases to Aiur, for example, U64 = [G; 8]


arthurpaulino (commented):
Looking mostly good!

Besides my comments below, I should say that the files where the new `U64` type has most value are the hash-related ones, `Blake3.lean` and `Sha256.lean`.

These `U64`s on the kernel types will be soon replaced by single `G`s anyways.

PR #366, opened 2026-04-07
Add non-tail match support (Yield + MatchContinue)

arthurpaulino:
Enable `let x = match y { 0 => return v, 1 => z }; use(x)` — match expressions in non-tail position, including branches with early returns.

## Bytecode

Two new Ctrl variants: Yield (terminates a non-returning branch, provides values for the merge) and MatchContinue (non-tail match with conditional-select merge and shared continuation block). MatchContinue carries sharedAuxiliaries and sharedLookups fields computed during lowering to align trace and constraint column positions.

## Lowering

findNonTailMatch walks through let chains (introduced by the simplification pass's match compiler) to detect embedded matches. yieldCtrl flag propagates through compile/addCase so non-escaping leaves produce Yield. The `!term.escapes` check is critical: the type checker erases the `.ret` wrapper, so escaping branches are distinguished only by the escapes flag on TypedTerm.

Degree tracking in CompilerState (opDegree mirrors opLayout's logic) provides exact operand degrees for computeSharedLayout, which also uses countReturns to subtract Return's pre-allocated slot-0 lookup from branch lookup counts.

## Constraints

get_block_selector recursively sums immediate children's selectors instead of the min_sel..max_sel range. This is the key fix for the selector double-counting bug: when a matchContinue is inside a tail match branch, the continuation's return selector fires alongside a yield selector. The recursive implementation excludes the continuation by only summing MatchContinue's case/default branches, not its continuation.

Ctrl::Match uses collect_branch_constraints and advances state.column/state.lookup to the branch max. Without this, when explicit branches are heavier than the default (e.g. branches with pointer-deref load ops), the outer scope sees only the default's lower column count, misaligning merge auxiliary positions.

Merge and linking constraints are gated by the parent selector (sel) for correctness when a matchContinue is in an inactive tail match branch — ungated constraints on shared auxiliary columns don't evaluate to 0 at the OOD evaluation point.

Per-block boolean constraints for all block selectors. Selector simplification pass (fixpoint substitution of sel_i = A).

## Execution and trace

Continuation stack with depth tracking in CallerState for correct cleanup on early returns across function boundaries.

populate_row returns Option<Vec<G>> to signal yield vs return. dispatch_branch helper shared by Match and MatchContinue. Trace advances index.auxiliary and index.lookup by sharedAuxiliaries and sharedLookups after the taken branch to skip the shared region.

### DIFF
--- a/Ix/Aiur/Compiler/Layout.lean
+++ b/Ix/Aiur/Compiler/Layout.lean
@@ -211,9 +211,7 @@ partial def blockLayout (block : Bytecode.Block) : LayoutM Unit := do
       maximalSharedData := maximalSharedData.maximals defaultBlockSharedData
       setDegrees degrees
     setSharedData maximalSharedData
-  | .return .. =>
-    bumpSelectors
-    bumpLookups
+  | .return .. => bumpSelectors
   | .yield .. => bumpSelectors
   | .matchContinue _ branches defaultBranch outputSize _sharedAux _sharedLookups continuation =>
     let initSharedData ← getSharedData
--- a/Ix/Aiur/Compiler/Layout.lean
+++ b/Ix/Aiur/Compiler/Layout.lean
@@ -214,6 +214,30 @@ partial def blockLayout (block : Bytecode.Block) : LayoutM Unit := do
   | .return .. =>
     bumpSelectors
     bumpLookups
+  | .yield .. => bumpSelectors
+  | .matchContinue _ branches defaultBranch outputSize _sharedAux _sharedLookups continuation =>
+    let initSharedData ← getSharedData
+    let mut maximalSharedData := initSharedData
+    let mut degrees ← getDegrees
+    for (_, block) in branches do
+      setSharedData initSharedData
+      blockLayout block
+      let blockSharedData ← getSharedData
+      maximalSharedData := maximalSharedData.maximals blockSharedData
+      setDegrees degrees
+    if let some defaultBlock := defaultBranch then
+      setSharedData initSharedData
+      bumpAuxiliaries branches.size
+      blockLayout defaultBlock
+      let defaultBlockSharedData ← getSharedData
+      maximalSharedData := maximalSharedData.maximals defaultBlockSharedData
+      setDegrees degrees
+    setSharedData maximalSharedData
+    -- Merge auxiliaries: one per output value
+    bumpAuxiliaries outputSize
+    pushDegrees (.replicate outputSize 1)
+    -- Continuation layout: additive (not maximals)
+    blockLayout continuation
 
 end Bytecode
 
--- a/Ix/Aiur/Compiler/Lower.lean
+++ b/Ix/Aiur/Compiler/Lower.lean
@@ -103,7 +103,7 @@ abbrev CompileM := EStateM String CompilerState
 
 /-- Compute the output degrees for a single-output op from its operand degrees.
     Multi-output ops (size > 1) always produce degree-1 values. -/
-private def opDegree (degrees : Array Nat) (op : Bytecode.Op) (size : Nat) : Array Nat :=
+private def pushOpDegree (degrees : Array Nat) (op : Bytecode.Op) (size : Nat) : Array Nat :=
   if size > 1 then
     (Array.replicate size 1).foldl (init := degrees) (·.push ·)
   else match op with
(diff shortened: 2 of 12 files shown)


Ix/Aiur/Compiler/Lower.lean
```diff
@@ -96,15 +96,32 @@ structure CompilerState where
...
   ops : Array Bytecode.Op
   selIdx : Bytecode.SelIdx
+  degrees : Array Nat
   deriving Inhabited
 
 abbrev CompileM := EStateM String CompilerState
 
+/-- Compute the output degrees for a single-output op from its operand degrees.
+    Multi-output ops (size > 1) always produce degree-1 values. -/
+private def opDegree (degrees : Array Nat) (op : Bytecode.Op) (size : Nat) : Array Nat :=
```
gabriel-barrett: Let's call this `pushOpDegree` for clarity

Ix/Aiur/Compiler/Lower.lean
```diff
@@ -321,38 +338,153 @@ partial def toIndex
...
+  go layoutMap bindings term := do
+    match term.inner with
+    | .match scrutinee cases => pure (some (term.typ, scrutinee, cases, bindings))
+    | .let (.var var) val bod => do
+      let val ← toIndex layoutMap bindings val
+      go layoutMap (bindings.insert var val) bod
+    | .let .wildcard val bod => do
+      let _ ← toIndex layoutMap bindings val
+      go layoutMap bindings bod
+    | _ => pure none
```
gabriel-barrett, line 367: There's also the possibility that a match statement is used inside an operation, like `(match ...) + 1`. Perhaps the compiler could be simplified if we did a "single-step" transformation before trying to compile to bytecode. Say, something like
```
(x + y) + z
```
would become
```
let tmp1 = x + y;
tmp1 + z
```
These temporary variables would be created using `Local.idx`

Ix/Aiur/Compiler/Layout.lean
```diff
@@ -214,6 +214,30 @@ partial def blockLayout (block : Bytecode.Block) : LayoutM Unit := do
   | .return .. =>
     bumpSelectors
     bumpLookups
```
gabriel-barrett: At the time I wrote this function I didn't think of `matchContinue` and early returns. I assumed there could only be one `return` in a path, since all `match` statements were in tail position. Now that we have early returns, a `return` can be sandwiched between two blocks, and it will be counted twice. All returns are shared, so they only take one lookup. Instead of bumping the lookup here, I would instead simply increase the lookup count by one when we compute the number of lookups by the function. That is: function lookups equal body (block) lookups plus 1

Ix/Aiur/Compiler/Lower.lean
```diff
@@ -321,38 +338,153 @@ partial def toIndex
...
+ (matchCases : Array (G × Bytecode.Block))
+ (default : Option Bytecode.Block)
+: CompileM (Nat × Nat) := do
+  let degrees := (← get).degrees
+  let initLS : Bytecode.LayoutMState :=
+    ⟨{ inputSize := 0, selectors := 0, auxiliaries := 0, lookups := 0 },
+     .empty, degrees⟩
+  let (sharedAux, sharedLookups) := matchCases.foldl (init := (0, 0)) fun (maxA, maxL) (_, block) =>
+    let (_, ls) := Bytecode.blockLayout block |>.run initLS
+    let effectiveLookups := ls.functionLayout.lookups - countReturns block
```
gabriel-barrett: I don't believe this is correct. You're subtracting the total number of returns instead of subtracting the number of wrongly counted returns (ones that are sandwiched). In any case, if you fix the computation of the layout, subtraction will not be needed.

src/aiur/constraints.rs
```diff
@@ -187,35 +260,167 @@ impl Ctrl {
...
+          state.constraints.zeros.push(sel.clone() * (merged.clone() - sum));
+          state.map.push((merged, 1));
         }
+
+        // Link continuation selector to the continuation block's selector
+        let cont_block_sel = continuation.get_block_selector(state);
+        state
+          .constraints
+          .zeros
+          .push(sel.clone() * (cont_block_sel - cont_sel.clone()));
```
gabriel-barrett: Replace this line with
```rust
          .push(cont_block_sel - cont_sel.clone());
```
These are structural constraints, they are true whether or not the path is taken

src/aiur/constraints.rs
gabriel-barrett: Since `Block::collect_constraints` now adds a boolean constraint to the selector, please remove this part.

Not that it matters too much, but perhaps we should create a `Block.collect_constraints_outermost` or some other name, that adds the boolean constraint on the selector, leaving `Block.collect_constraints` without this constraint. The reason is that we only need to constraint the outermost block, which are the initial function body and all continuation blocks.

gabriel-barrett (approved):

PR #370, opened 2026-04-08
Aiur interpreter

gabriel-barrett:
This PR implements an interpreter of Aiur terms, used for debugging. Features
- [x] Early return
- [x] Stack trace
- [x] IO buffer
- [x] Pointer caching
- [x] Function call caching


Ix/Aiur/Interpret.lean
```diff
@@ -130,10 +130,15 @@ instance : ToString Interrupt where
...
 structure InterpState where
-  heap : Heap
-  ioBuffer : IOBuffer
+  heap       : Heap
+  ioBuffer   : IOBuffer
+  /-- Memoized results of pure function calls: (function name, args) → result. -/
+  callCache  : Std.HashMap (Global × List Value) Value := {}
+  /-- Inverse of the heap: value → existing pointer index.
+      Ensures each distinct value is stored at most once. -/
+  storeCache : Std.HashMap Value Nat := {}
```
arthurpaulino: I think instead of `heap` and `storeCache` we can use an `IndexMap` (already defined and available in `Ix/IndexMap.lean`) to fulfill both roles.

samuelburnham (approved):

PR #374, opened 2026-04-10
feat: Benchmark throughput, R^2, outlier variance

samuelburnham:
This PR makes several improvements to the Benchmark.lean library:

- Adds throughput tracking per benchmark, which required creating a `BgroupM` state monad so we can thread the input sizes through to each benchmark. This also cleans up the UX so you can run any custom logic around/between bench calls in a `do` block instead of pushing each bench to an array.
- Adds R^2 regression analysis and outlier variance detection based on https://github.com/haskell/criterion, which isn't present in https://github.com/criterion-rs/criterion.rs but gives additional metrics for user confidence.
- Decreases the default bootstrap sample count from 100k (used in criterion.rs) to 10k. 10k resamples is sufficient for stable confidence interval estimates at the default sample size of 100 or similar, and the 10x reduction in analysis time (5s->.5s, where criterion.rs is also about .5s) is a meaningful UX improvement.
- Adds an `Ansi.lean` color-printing module, and fully replicates the pretty-printed console output of criterion.rs.
- Factors out display/reporting logic from `Bench.lean` into `Report.lean`.

### DIFF
--- a/Benchmarks/Aiur.lean
+++ b/Benchmarks/Aiur.lean
@@ -122,8 +122,7 @@ def verifyBench : IO $ Array BenchReport := do
   bgroup "nat_fib" {} do
     bench "verify fib 10" (Aiur.AiurSystem.verify system friParameters claim) proof
 
-def main (args : List String) : IO Unit := do
-  setBenchArgs args
+def main : IO Unit := do
   let _ ← toplevelBench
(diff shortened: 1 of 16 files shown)


arthurpaulino (approved):
LGTM!

PR #378, opened 2026-04-14
chore: Update iroh and rand

samuelburnham:
- Updates iroh to v0.97 and fixes the Lean `target` to correctly build `ix_rs_net` with the Rust `net` feature. This method is the same as `ix_rs_test`, which adds a couple of extra seconds to the Lean build when the Rust changes since it has to first build `ix_rs` with default features, then rebuild with `--features <net|test-ffi>`. It's an incremental rebuild though so not too bad, especially since we usually edit/recompile either Lean or Rust at a given time and not both, thereby avoiding the duplicate build in the custom target.
- Updates rand to v0.10.1 for ix tests and replaces iroh rand with getrandom, which resolves `RUSTSEC-2026-0097`
- Removes stale advisory ignores from `deny.toml`


arthurpaulino (approved):
Thanks! The iroh dependency (tree) really needed an update!

PR #380, opened 2026-04-14
chore: Update lean-ffi

samuelburnham:
Update `lean-ffi` which now has more ergonomic methods for working with `LeanCtor` scalar types.

After https://github.com/argumentcomputer/lean-ffi/pull/8


arthurpaulino (approved):
These should be mostly mechanical changes to adapt to the new API.

PR #388, opened 2026-04-17
feat: Polymorphic benchmark `bgroup`

samuelburnham:
- Wraps `Benchmarkable` functions in an `IO Unit` closure that allows mixing different types of benchmarks in the same `bgroup`
- Adds `benchStep`, `benchStepIO` and `benchStepE` methods that return the output of the last bench iteration, which allows threading into the inputs of a pipelined benchmark. Updates the Aiur benchmark accordingly, so the benches now read like a normal program
- Adds an `E2E` metric that adds all sampled mean and one-shot results into a final end-to-end wall-clock time, which is then printed and/or included in the Markdown report as its own row. Code blocks containing benchmarks can skip this calculation with the `skipE2E` function, and toggle it back on with `countInE2E`.
- Fixes report pretty-printing for the `bench-pr.yml` workflow

### DIFF
--- a/.github/workflows/bench-pr.yml
+++ b/.github/workflows/bench-pr.yml
@@ -130,9 +130,8 @@ jobs:
           {
             echo '## Benchmark for `${{ matrix.bench }}` at ${{ env.COMMIT_LINK }}';
             echo "";
-            for file in benchmark-report-*.md; do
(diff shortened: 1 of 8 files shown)


arthurpaulino (approved):
👏🏼

PR #417, opened 2026-05-19
Intern addresses as `&[G; 32]` to cut kernel FFT cost

arthurpaulino:
Replace the by-value `[G; 32]` address type with `type Addr = &[G; 32]`. Aiur `store` content-dedups structurally equal arrays to one canonical pointer, so address equality reduces to pointer subtraction and every function carrying addresses drops 31 columns per address argument.

`lake exe kernel String.Internal.append`: total FFT 2515875651 ->
2152947064 (-14.4%), total width 48210 -> 37903. ixvm test suite: 283 pass, 0 fail.


Ix/IxVM/Ingress.lean
```diff
@@ -1218,7 +1219,7 @@ def ingress := ⟦
         -- Check if this address has constant data in IOBuffer.
         -- io_get_info is unconstrained; the prover provides (0, 0) for blob addresses.
         -- Soundness: if the prover lies and skips a real constant, type checking will fail.
-        let (_, len) = io_get_info(addr);
+        let (_, len) = io_get_info(load(addr));
```
gabriel-barrett, line 1222: I believe this is doing a lookup and loading the addresses into columns JUST to get a hint. The hint should somehow come from the pointer alone. Is this possible? Perhaps that will need a new `io_load_info` operation or something?

arthurpaulino: Nice idea

Ix/IxVM/Ingress.lean
gabriel-barrett, line 74: Again this seems to be doing a bunch of work just to get a hint. This whole function seems a bit confusing because it is mostly hint. Is it being called unconstrained?

gabriel-barrett (approved):

PR #427, opened 2026-05-23
feat: Bench Aiur kernel on any def in LEAN_PATH

samuelburnham:


Benchmarks/Typecheck.lean
```diff
@@ -19,25 +27,34 @@ def friParameters : Aiur.FriParameters := {
   queryProofOfWorkBits := 0
 }
 
-def main : IO Unit := do
+def parseName (s : String) : Lean.Name :=
+  s.splitOn "." |>.foldl (init := .anonymous) Lean.Name.mkStr
```
arthurpaulino, line 31: This doesn't work very well, because it can't handle names such as `Foo.0.Bar`.

It's already solved in Check.lean.

samuelburnham:
Superseded by #449

PR #433, opened 2026-06-02
IxVM: context-suffix memo keys for closed terms + drop dead KValNode

arthurpaulino:
Two changes to the Aiur kernel core, both behavior-preserving.

Closed-term context normalization in whnf / k_infer / k_is_def_eq: a closed term (`expr_lbr(e) == 0`) reads nothing from the local binder-type context `types` — its only reader, `types_lookup`, is reached via a `BVar` lookup, which a closed term has none of. So the result is independent of `types`; collapse it to `Nil`. The same closed subterm appearing under different binder depths then shares one memo key `(e, Nil, top, addrs)` instead of missing on a per-depth `types`, so Aiur's content-memoization reuses it. Open terms keep their real context. Sound: a closed term reduces and infers only over closed subterms.

Remove the dead `KValNode`/`KVal`/`KValEnv` NbE value domain: defined but referenced nowhere (the live kernel runs on de-Bruijn `KExpr`); vestigial from an abandoned NbE direction.

Measured with `ix check --ixe arena.ixe --claim <CheckEnv-digest>` over the TutorialDefs arena (694 consts): 7,159,028,984 -> 7,151,454,776 FFT (-0.11%). `lake test -- --ignored ixvm`: 297 pass, 0 fail. (The closed-term win is larger on a faster ingress base, where the kernel core is a bigger share of total.)


arthurpaulino:
Superseded by #441

PR #438, opened 2026-06-10
IxVM kernel: cut Aiur FFT cost on reduction- and check-heavy shards

arthurpaulino:
## Summary

Thirteen commits in the Aiur IxVM kernel: ten optimization areas — each
targeting a specific hotspot found by profiling per-function FFT cost on real
shards or single constants of a compiled `Init` environment — plus a CLI
ergonomics fix for resolving private `Lean.Name` arguments. No semantics change:
every shard still checks without error, and the full `lake test -- --ignored
ixvm` suite passes (297 tests) at every step.

Cumulative effect on a reduction-heavy shard (Init shard 9, one owned constant,
deep reduction): **45.5G → 16.6G total FFT (−64%)**. The structural / keyed-lookup
areas (1–5) account for 45.5G → 37.8G; the Nat-reduction area (6) takes it the
rest of the way by collapsing two symbolic-arithmetic blowups from O(n) / O(n²)
to O(1). That same area makes a previously intractable proof class — the UTF-8
codec lemma `ByteArray.utf8DecodeChar?_utf8EncodeChar_append`, whose checking
spawns deep `Nat.add x 0xC0` succ-towers and `c >>> 6/12/18` division-towers —
reduce in constant rather than tower-depth cost. The deserialization (7),
memoization (8), substitution (9), and primitive-gauntlet (10) areas then cut
ingress and reduction cost across all consts.

Those last areas make another previously intractable proof checkable:
`Vector.extract_append._proof_1` (an `omega`-backed `Vector.extract`/`++` lemma
whose check is ~50% beta-substitution) went from **OOM (>14G RAM)** to 56.6G FFT
(area 8, which lets it complete at all) to 38.3G (area 9, −32%) to **28.0G**
(area 10, a further −27%) — **−50%** from the first completing run.

## Method

Built a `.ixe` of the `Init` closure, profiled and sharded it
(`ix profile` → `ix shard --shards 64`), then ran `ix check` per shard or per
constant and read the per-function FFT table (`--stats-out`). Shard 10
(check-heavy: lookups, address compares, deserialize) and shard 9
(reduction-heavy: whnf, substitution, spine collection) exercise different
profiles; small hand-built or hand-picked reproducers (`x + D = …`, `x >>> D = …`,
`Int.emod_emod_of_dvd`) isolate and scale individual blowups.

Aiur cost model behind the changes: FFT ≈ `per-function width × row height`,
charged at the function's **max-arm width on every row**. Levers: (a) factor a
wide/rare arm into its own circuit; (b) cut row count via better memoization;
(c) replace linear scans with keyed lookups; (d) keep symbolic primitives
compact so they never materialize a tower.

## Areas

**1. Cheaper `Expr.Share` / universe lookups** — commit `3774860`.
The sharing-table and universe-array lookups walked the list with
`list_lookup_u64` (per-step U64 predecessor). Switched to a field index and, for
`Expr.Share`, to `list_drop` — which returns the sublist *pointer*, so repeated
lookups dedup in the memo cache. Shard 10: total FFT 3.89G → 2.11G; the merged
pointer-list walk fell from 1.93G to 5.8M.

**2. Hot/cold split of `address_eq`** — commit `69b6936`.
The primitive-dispatch gauntlet compares each `Const` head against ~50 hardcoded
addresses, making `address_eq` (a full 32-limb compare, width 109) the hottest
circuit. Almost every comparison differs at limb 0. Split it: `address_eq`
compares limb 0 only and delegates the rest to `address_eq_tail`, whose height is
just the rare limb-0 matches. (A nested match in the same function does nothing —
Aiur charges full width per row regardless of arm; the split must be a function
boundary.) Shard 10: width 109 → 80, 509.8M → 374.1M; total 2.106G → 1.973G.

**3. Hot/cold split of `whnf_with_spine`** — commit `0e30d78`.
On reduction-heavy consts this is the single hottest circuit (width 89 × 3.36M
rows), its width set by two arms a minority of rows reach: Const-head and
Proj-head dispatch. Factored both into `whnf_const_head` / `whnf_proj_head`. The
Proj arm is the jackpot — 3.4K rows but nearly half the width. Shard 9: width
89 → 34, 6.48G → 2.48G.

**4. Non-tail, single-definition `collect_spine`** — commit `5b7fabc`.
`collect_spine` peeled the App chain with a tail-recursive accumulator. Tail
recursion buys nothing in Aiur (stack depth is free), and the accumulator landed
in the memo key, blocking reuse — 0 cache hits across 5.38M rows. Rewrote it
non-tail, keyed on the expression alone (2.19M hits now), and merged the two
verbatim copies into one definition in `kernelTypes`. Shard 9: 2.65G → 0.96G.

**5. Keyed skip-set for sharded check** — commit `a52967f`.
The largest `address_eq` source on a *sharded* check was `addr_in_list`, an O(N)
linear scan of the assumption-leaf list per checked const (1.81M calls on shard
9). Replaced with an `RBTreeMap` skip-set keyed on the first 4 address bytes:
build once, membership = one rbtree lookup + one confirming `address_eq`. Shard 9:
`address_eq` 2.07M rows / 3.48G → 274K rows / 0.40G; total 41.7G → 37.8G.

**6. Compact symbolic Nat reduction** — commits `dbc4177`, `5dcab7f`, `f7cfe23`.
`Nat` primitives applied to (symbolic base, literal) delta-unfolded into towers
the kernel then re-walked: `Nat.add x 0xC0` → `succ^192(x)`, and `Nat.div x 2`
(what `x >>> k` unfolds to, k times) → an exploding division algorithm. Three
coordinated commits keep them compact:
- `dbc4177`: `try_nat_linear_rec` reduces `Nat.rec base (succ-step) (Lit n)` with
  a *symbolic* base directly to `Nat.add base (Lit n)` instead of n iota steps.
- `5dcab7f`: whnf (`try_nat_offset_stuck`) leaves `Nat.add base (Lit n)` stuck as
  a compact offset; def-eq (`nat_offset_of` + `try_def_eq_nat`) decomposes each
  side to `base + KLimbs` and decides `base ≟ base` on equal offsets.
- `f7cfe23`: generalizes the stuck rule to `Nat.div` / `Nat.mod base (Lit k)`
  (k ≥ 2), so symbolic `>>>` reduces to nested-stuck-div in O(1).

Benchmarks: `x + D = Nat.rec x succ-step D` goes O(D) → depth-independent
(328M → 28.5M at D=512, −91%); `x >>> 18 = (x >>> 9) >>> 9` goes 15.65G → 455M
(−97%), matching `x >>> 6`. All magnitudes stay `KLimbs` (no Goldilocks
collapse); only de Bruijn / `top`-position indices are `G`.

**7. Narrow Ixon expr deserialization** — commit `fa3447e`.
On large-closure consts, ingress deserialization dominates: the two hottest
parsers are `get_expr` and `get_app_telescope`, each carrying width set by a
single arm. Split `get_expr`'s `Let` arm (three sub-`get_expr` calls + triple
`store`) into `get_expr_let`, dropping `get_expr` width 102 → 75. Changed
`get_app_telescope`'s `func: Expr` parameter to `func: &Expr` (load inside),
dropping width 95 → 78. Ingress profile (2000-const UTF-8 closure, run_check
load-only): `get_expr` 766M → 563M, `get_app_telescope` 477M → 392M, total
ingress 3.13G → 2.85G.

**8. lbr-trimmed context keys for whnf / infer / def-eq memoization** — commit
`d96c738`.
Aiur auto-memoizes `whnf` / `k_infer` / `k_is_def_eq` on the full
`(expr, types, …)` argument tuple, so the *same* closed reduction gets a fresh
memo key at every binder depth and misses. Mirror Rust's
`whnf_key` / `infer_key` / `def_eq_ctx_key` (`src/ix/kernel/tc.rs`): trim `types`
to the suffix reachable from the expression's loose-bvar range (`ctx_trim` /
`ctx_reachable`, a fixpoint that closes over kept frames' own ranges) before the
memoized call. A closed subterm (`lbr == 0`) keys to the empty context and shares
its result across every depth. Eager substitution is unchanged (matches the
Lean / Rust reference); the win is purely cache keying. Fast paths
(closed → `Nil`, full-reach → as-is) keep the boundary cheap.
FFT (`ix check <const> --stats-out`): `Nat.add_comm` 58.28M → 57.90M (−0.66%);
`Int.emod_emod_of_dvd` 5.090G → 4.691G (−7.85%, inference / omega-heavy); Init
shard 10 4.738G → 4.721G (−0.37%). Net-positive everywhere measured.

**9. Multi-arg beta with a narrow substitution walk** — commit `3203691`.
`whnf_apply_beta` substituted one spine argument per `expr_inst1`, re-walking
the (shrinking) body for each, so a K-argument application `(λλλ. body) x y z`
walked the body K times. Substitution (`expr_inst1` + `expr_inst1_walk`) was
**~53%** of FFT on `omega`-heavy proofs. Mirror Rust's multi-arg beta
(`src/ix/kernel/whnf.rs:541-567`): `peel_beta` consumes the whole telescope and
`expr_inst_many` (= `simul_subst`) substitutes all consumed args in **one** walk.
Single-arg betas stay on the cheap `expr_inst1` path, so short telescopes don't
pay the list overhead — the reason an earlier *unconditional* `simul_subst`
regressed and was reverted. A hot/cold split factors the multi-arg walk's `BVar`
arm (`list_length` / `list_lookup` / `expr_lift`) into `expr_inst_many_bvar`,
keeping the hot App/Lam walk narrow (width 82 → ~53, the same pattern as areas
2–3). FFT (`ix check <const> --stats-out`):
`…Vector.extract_append._proof_1` 56.64G → 38.31G (−32.4%); `Nat.add_comm`
57.90M → 56.17M; Init shard 10 4.721G → 4.668G (−1.1%); shard 9 16.346G →
16.314G.

**10. Gate the primitive-op gauntlet behind a memoized classifier** — commit
`87cb8b0`.
`whnf_const_head` ran five `try_*` dispatchers (nat / str / bitvec / native /
decidable) in sequence on **every** `Const` head, each comparing the head
address against its family's primitive addresses — wasted on the common
non-primitive head. Worse, several `try_*` **speculatively WHNF their arguments**
(to test for literal / decidable operands), so a non-primitive head still
triggered deep argument reductions that were then thrown away. Factor the chain
into `try_address_primitives` and gate it behind `prim_any_addr`: a sum of the 43
head-dispatched `address_eq` checks, **memoized per head address** (computed once
per const, not per whnf). A `0` means no `try_*` can match, so dispatch jumps
straight to projection-definition / delta. Validated by a *differential* assert
(`prim_any_addr` must be 1 whenever the gauntlet actually matched) run over the
suite + `Int.emod_emod_of_dvd` + `Vector._proof_1`; it caught two missing
addresses (`string_utf8_byte_size`, `punit_size_of_1`), then passed across
millions of reductions — assert removed afterward. FFT (`ix check <const>`):
`…Vector.extract_append._proof_1` 38.31G → 28.02G (−26.8%, the skipped
speculative arg-WHNF); `Int.emod_emod_of_dvd` 4.69G → 3.97G (−15%);
`Nat.add_comm` 56.17M → 56.06M; Init shard 9 16.314G → 16.279G; shard 10
4.668G → 4.659G.

## Tooling

**Resolve private `Lean.Name` args in check/prove** — commit `6ec124a`.
`parseName` (naive dot-split; an all-digit segment → `Name.mkNum`) can't rebuild
a private name like `_private.Init.Data.Vector.Extract.0.…_proof_1` from its
display string, so `ix check` / `ix prove` reported it "not found". Added
`toString` fallbacks — `resolveName` (compiled-in env) and `resolveIxeAddr`
(`.ixe` env) — that match the arg against each constant's displayed name when the
direct lookup misses. Covers both commands (shared `forEachClaim`), across the
compiled-in and `--ixe` paths. Tooling only; no kernel change.

## Soundness

- Areas 1, 3, 4, 7 are structural / refactor rewrites with identical results
  (each shard still checks).
- Area 2: `address_eq` returns the same value as a full 32-limb compare — one
  differing limb proves inequality.
- Area 5: key collisions overwrite, sound because the only effect is a *missed
  skip* (a frontier const gets rechecked, never wrongly trusted); the confirming
  `address_eq` rules out false positives; the 4-byte key makes collisions
  negligible regardless.
- Area 6: keeping `Nat.add / div / mod symbolic literal` stuck is the correct
  normal form (these are irreducible for a symbolic base). `try_def_eq_nat`
  returns `eq = 1` only when offsets are equal and bases are def-eq (`+ k` is
  injective); otherwise it falls back to the generic path. `BAD` negative cases
  still get rejected (no false accept), and recursor-on-symbolic-offset reduces
  correctly.
- Area 8: trimming `types` to the loose-bvar-reachable suffix changes only the
  *memo key*, not the computation — a closed expr never consults the trimmed
  frames, and `ctx_reachable`'s fixpoint keeps every frame a kept frame's stored
  type can reach. Eager substitution and all reduction logic are untouched.
- Area 9: `expr_inst_many` is the standard simultaneous de Bruijn substitution
  (`simul_subst`); `peel_beta` consumes exactly one argument per `Lam`, so
  `(λλλ. body) x y z` and three sequential single substitutions produce the same
  term. The hot/cold split only relocates the `BVar` arm — identical result.
  Eager substitution semantics are unchanged (matches the Lean / Rust reference).
- Area 10: gating is sound iff `prim_any_addr` is `1` whenever the gauntlet would
  reduce — skipping it then only elides work that returns "no match". A
  differential `assert` checked exactly this invariant on the suite plus the two
  heavy consts (millions of reductions across every primitive family) before the
  gate was trusted; the assert caught the only two omissions and then ran clean.
- Whole suite: the full `ixvm` suite passes (297 tests) at every commit.

### DIFF
--- a/Ix/Cli/CheckCmd.lean
+++ b/Ix/Cli/CheckCmd.lean
@@ -45,6 +45,16 @@ def parseName (arg : String) : Lean.Name :=
       | some n => Lean.Name.mkNum acc n
       | none   => Lean.Name.mkStr acc s
 
+/-- Resolve a CLI name argument against the env. `parseName` can't rebuild
+    private names (`_private.M.0.foo`) — the marker/scope-index components
+    don't round-trip through naive dot-splitting. So if the parsed name
+    isn't present, fall back to matching the arg against each constant's
+    `toString` (the displayed form the user copied). -/
+def resolveName (env : Lean.Environment) (arg : String) : Option Lean.Name :=
+  let parsed := parseName arg
+  if env.constants.contains parsed then some parsed
+  else (env.constants.toList.find? (fun (k, _) => toString k == arg)).map (·.1)
+
 def addrOfHex! (label : String) (s : String) : IO Address := do
   match Address.fromString s with
   | some a => pure a
@@ -87,6 +97,16 @@ partial def ixNameToLeanName : Ix.Name → Lean.Name
   | .anonymous _ => .anonymous
   | .str p s _ => .str (ixNameToLeanName p) s
   | .num p n _ => .num (ixNameToLeanName p) n
+
+/-- Resolve a CLI name argument against a `.ixe` env's named map. Like
+    `resolveName` for the compiled-in env: `parseName` can't rebuild private
+    names, so when the direct lookup misses, fall back to matching the arg
+    against each named constant's displayed `toString`. -/
+def resolveIxeAddr (ixonEnv : Ixon.Env) (arg : String) : Option Address :=
+  match ixonEnv.getAddr? (Ix.Name.fromLeanName (parseName arg)) with
+  | some a => some a
+  | none =>
+    (ixonEnv.named.toList.find? (fun (k, _) => toString (ixNameToLeanName k) == arg)).map (·.2.addr)
 
 /-- Build a `ClaimWitness` for the `verify_claim` entrypoint against
     `Ix.Claim.check addr none` (full transitive typecheck of the
@@ -189,14 +209,13 @@ def forEachClaim
           if !keepGoing then break
     else
       for arg in names do
-        let name := parseName arg
-        match ixonEnv.getAddr? (Ix.Name.fromLeanName name) with
+        match resolveIxeAddr ixonEnv arg with
         | none =>
-          IO.eprintln s!"{name} not found in {path}"
-          failures := failures.push (toString name)
+          IO.eprintln s!"{arg} not found in {path}"
+          failures := failures.push arg
           if !keepGoing then break
         | some addr =>
-          let label := toString name
+          let label := arg
           let claim := Ix.Claim.check addr none
           let witness ← mkWitness addr ixonEnv
           if (← runOne claim witness label) ≠ 0 then
@@ -224,17 +243,18 @@ def forEachClaim
           if !keepGoing then break
     else
       for arg in names do
-        let name := parseName arg
-        if !env.constants.contains name then
-          IO.eprintln s!"{name} not found"
-          failures := failures.push (toString name)
+        match resolveName env arg with
+        | none =>
+          IO.eprintln s!"{arg} not found"
+          failures := failures.push arg
           if !keepGoing then break
           else continue
-        let label := toString name
-        let (claim, witness) ← buildOne name
-        if (← runOne claim witness label) ≠ 0 then
-          failures := failures.push label
-          if !keepGoing then break
+        | some name =>
+          let label := toString name
+          let (claim, witness) ← buildOne name
+          if (← runOne claim witness label) ≠ 0 then
+            failures := failures.push label
+            if !keepGoing then break
 
   if failures.isEmpty then pure 0
   else
--- a/Ix/IxVM/Convert.lean
+++ b/Ix/IxVM/Convert.lean
@@ -68,7 +68,9 @@ def convert := ⟦
     match load(idxs) {
       ListNode.Nil => store(ListNode.Nil),
       ListNode.Cons(idx, rest) =>
-        let u = load(list_lookup_u64(univs, idx));
+        -- universe indices are small; walk with a field index (cheap per-step
+        -- field sub) instead of `list_lookup_u64`'s per-step U64 predecessor.
+        let u = load(list_lookup(univs, flatten_u64(idx)));
         store(ListNode.Cons(store(convert_univ(u)), convert_univ_idxs(rest, univs))),
     }
   }
@@ -145,7 +147,9 @@ def convert := ⟦
   ) -> KExpr {
     match load(e) {
       Expr.Srt(univ_idx) =>
-        let u = load(list_lookup_u64(univs, univ_idx));
+        -- field-indexed walk (see `convert_univ_idxs`): avoids the per-step
+        -- U64 predecessor of `list_lookup_u64` on this hot universe lookup.
+        let u = load(list_lookup(univs, flatten_u64(univ_idx)));
         store(KExprNode.Srt(store(convert_univ(u)))),
 
       Expr.Var(idx) =>
@@ -199,7 +203,8 @@ def convert := ⟦
           convert_expr(body, sharing, ref_idxs, recur_idxs, lit_blobs, univs))),
 
       Expr.Share(idx) =>
-        convert_expr(list_lookup_u64(sharing, idx), sharing, ref_idxs, recur_idxs, lit_blobs, univs),
+        let ListNode.Cons(e, _) = load(list_drop(sharing, flatten_u64(idx)));
+        convert_expr(e, sharing, ref_idxs, recur_idxs, lit_blobs, univs),
     }
   }
 
--- a/Ix/IxVM/Ingress.lean
+++ b/Ix/IxVM/Ingress.lean
@@ -53,17 +53,23 @@ def ingress := ⟦
     bytes
   }
 
-  -- Compare two 32-byte addresses for equality
-  fn address_eq(a: Addr, b: Addr) -> G {
-    let [a0, a1, a2, a3, a4, a5, a6, a7,
+  -- Compare two 32-byte addresses for equality.
+  --
+  -- Cold path: limb 0 already matched, compare the remaining 31 limbs.
+  -- Factored into its own function so it forms a separate circuit whose height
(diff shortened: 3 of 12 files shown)


johnchandlerburnham:
Are any of these improvements able to be ported to the Rust kernel implementation?

Also, can we run execution out-of-circuit on Init, Lean & Mathlib?

johnchandlerburnham (approved):

PR #440, opened 2026-06-11
Recursive verifier

gabriel-barrett:
Implements the recursive verifier, that is, multi-stark in Aiur, including Plonky3's PCS verifier.
Currently it assumes values of a few parameters:
1. `capHeight = 0`, meaning commitments are a single digest. Cap height decides how deep into the merkle tree the commitment is. A greater cap height means the commitment themselves are larger, but it saves on the openings.
2. `maxLogArity = 1`, meaning FRI uses arity 2 fold.
3. `queryProofOfWork = 0`, meaning there's no proof of work before the sampling of query indices

Eventually we could generalize it to accept other values for `capHeight` and `queryProofOfWork`. This is mostly for proof size. Regarding `maxLogArity`, it might not be worth it generalizing it, since arity 2 is pretty standard


RecursiveVerifier.lean
```diff
@@ -0,0 +1,201 @@
+import Ix.Aiur.Meta
```
arthurpaulino: This file seems to be a smoke test for the recursive verifier.
I think testing should be properly integrated in the test suite.

Cargo.toml
```diff
@@ -13,7 +13,7 @@ itertools = "0.14.0"
 indexmap = { version = "2", features = ["rayon"] }
 lean-ffi = { git = "https://github.com/argumentcomputer/lean-ffi", rev = "cc98ebf67bf453ac3827cb767f78b13ea674dd6a" }
 mimalloc = { version = "0.1", default-features = false }
-multi-stark = { git = "https://github.com/argumentcomputer/multi-stark.git", rev = "9ecab51d553445c0cc7b571af00a76b8a83a6f8c" }
+multi-stark = { path = "multi-stark" }
```
arthurpaulino: This change needs to be undone.

Ix/MultiStark/Deserialize.lean
```diff
@@ -0,0 +1,406 @@
...
+  -- `Commitment = MerkleCap<..>`: only the `cap: Vec<Digest>` is on the wire.
+  type MerkleCap = List‹Digest›
+
+  -- `OpenedValuesForRound<ExtVal> = Vec<Vec<Vec<ExtVal>>>`.
+  type OpenedRound = List‹List‹List‹Ext›››
+
+  -- The three trace/quotient commitments at the head of a `Proof`.
+  enum Commitments {
+    Mk(MerkleCap, MerkleCap, MerkleCap)
+  }
```
arthurpaulino, line 58: Seeing these single-constructor enums gave me the idea: they don't need tags in the bytecode and should be treated as simple tuples

arthurpaulino (dismissed):
Unblocking

arthurpaulino (approved):

PR #441, opened 2026-06-12
perf(aiur): Canonical levels, memoized dispatch, flat query record

samuelburnham:
## Description
  Kernel and record optimizations. Measured against current `main` (same machine,
  `lake exe ix check --ixe` over full `Init`):

  - `Int16.instRxcHasSize_eq`: **does not complete on main** (killed at 120 GB RSS,
    still growing) → **2m00s / 16.8 GB / 202.6 B FFT cost**.
  - Execution (witness generation): **~3× faster, ~2.5× less RAM** on every workload.
  - FFT cost on completing constants: −0.2% to −2.0% on real `Init` constants,
    −5.5% on a literal-fold synthetic.

  Units: M/B = millions/billions of FFT cost units; GB = memory.

  Commits, oldest first:

  **Aiur: flat append-only QueryMap + env-gated record-size stats**
  Replaces the per-circuit `IndexMap<Vec<G>, QueryResult>` record with flat per-circuit
  arenas (contiguous key/output/multiplicity vectors + a compact hash index). Same
  entries in ~54 bytes instead of ~133; circuit stats bit-identical. ~2.5× less
  execution RAM, ~3× faster execution. Also adds the opt-in `IX_AIUR_QUERY_STATS=1`
  record dumps used to find the kernel pathologies below.

  **IxVM: inline lbr guards on expr_inst1_walk App children**
  Substitution recorded a memo row even for subtrees with no bound variable. The walk
  now skips no-op subtrees inline (application arms only — guarding every arm widens
  the hot circuit past the savings). −68% rows in the substitution dispatch circuit on
  instantiation-heavy checks.

  **IxVM: drop dead KValNode/KVal/KValEnv**
  Removes an unreferenced normalization-by-evaluation value domain. Cleanup only.

  **IxVM: memoized prim_family dispatch + width-safe offset-stuck placement**
  Const-head whnf ran five primitive reducers in sequence on every head. A classifier
  keyed on the constant's address (memoized once per constant per run) names the one
  family that could fire; dispatch calls at most that reducer. Also moves the
  symbolic-Nat offset-stuck check off the hot path into the Nat reducer's miss path as
  a cold function. Largest single dispatch win on the suite; supersedes #438's boolean
  gauntlet gate (audited: identical effective head coverage, and primitive heads skip
  the gauntlet walk too).

  **IxVM: canonical level normalization for level_equal / level_leq**
  Replaces the recursive universe-level comparison — exponential in the number of level
  params from its two-way Max/IMax splits — with normalize-then-compare canonical forms
  (as in the Rust kernel and Lean4Lean). Each distinct level normalizes once per run.
  This is the completability fix: Int16 went from non-completing (level circuits were
  93% of its record) to ~2 minutes. The rest of the suite is unaffected.

  **Ix CLI: refs-of subcommand**
  Prints a constant's direct references in the format `ix tree canonical` expects, so a
  single-constant assumption-gated check is a three-command pipeline. Tooling only.

  **IxVM: idx-keyed projection-definition classification**
  "Is this definition a projection wrapper" was decided per call site with the argument
  spine in the memo key; the answer depends only on the constant. Now keyed on the
  constant index (one row per constant per run). Drops that circuit from 12.7 M record
  entries to a few hundred on UTF-8-class checks.

  **IxVM: port jcb/fixes H-14 — ptr_val skip map + lockstep addr cursor**
  Two sharded-check scaling fixes (from jcb/fixes, John C. Burnham): the assumption-leaf
  skip set keys on interned pointers instead of address prefixes, and the constant
  iterator walks the address list in lockstep instead of re-indexing per constant —
  removing a standalone O(closure²). FFT-neutral on single-constant checks.
## Benchmarks
  | constant | source | main | sb/aiur-perf | Δ |
  |---|---|---|---|---|
  | `_private…SInt.0.Int16.instRxcHasSize_eq` | Init | **DNF** (killed: 120 GB RSS, 8m45s, non-converging) | **202.6 B** (2m00s, 16.8 GB) | **DNF → completes** |
  | `_private…Vector.extract_append._proof_1` | Init | 28.02 B | 27.46 B | −2.0% |
  | `_private…Vector.extract_append._proof_2` | Init | 28.03 B | 27.47 B | −2.0% |
  | `Nat.fold_congr` | Init | 11.00 B | 10.92 B | −0.7% |
  | `Int.emod_emod_of_dvd` | Init | 3.97 B | 3.94 B | −0.7% |
  | `Nat.gcd_comm` | Init | 1,998.3 M | 1,994.7 M | −0.2% |
  | `Nat.shiftRight_eq_div_pow` | Init | 1.74 B | 1.74 B | −0.4% |
  | `mergeSortBench` | synthetic | 1,442.4 M | 1,438.0 M | −0.3% |
  | `String.append` | Init | 1,117.1 M | 1,117.3 M | +0.0% |
  | `foldAdd_2000` | synthetic | 265.0 M | 250.5 M | **−5.5%** |
  | `List.length_append` | Init | 112.3 M | 111.6 M | −0.6% |
  | `Nat.add_comm` | Init | 56.1 M | 55.4 M | −1.2% |
  | `ByteArray.utf8DecodeChar?_utf8EncodeChar_append` | Init | DNF | DNF — execution RAM 246 GB → 104 GB at equal ops | RAM only |


samuelburnham:
Superseded by #444 and #447

PR #445, opened 2026-06-16
perf(ix): lazy, zero-copy anon env load for the Aiur check path

samuelburnham:
`ix check --ixe` deserialized the whole env eagerly: the Rust FFI materialized every constant and the Lean side built an object graph for all of them up front, even to check one lemma. On mathlib.ixe (2.97 GB) that was ~104 GB peak RSS / ~159 s just to load.

Make the Lean `Ixon.Env` constants lazy and load them without copying:

- `LazyConstant` is now an offset window `(buf, off, len)` into a shared backing buffer (+ optional materialized cache for the build path), parsed on demand. Only the checked closure is ever materialized.
- New `rs_de_env_lazy` FFI: `Env::parse_lazy_index` walks the env once (reusing the existing parser, so every metadata variant incl. `CallSite` is handled) and returns per-const `(addr, offset, len)` windows into the buffer Lean passed in, plus `name -> addr` and per-`Defn` reducibility hints. No constant body is parsed or copied at load time. `deEnvAnon` reconstructs an `Env` of byte-window `LazyConstant`s over that same buffer.
- This is anon-aligned with the circuit: binder metadata (the `ExprMetaArena`) is parsed-and-discarded, since the typecheck circuit only consumes anonymous constants, blobs, and the per-`Defn` hint. The CLI's `name -> addr` map is kept so targets can still be named.
- Pure-Lean `getEnv`/`deEnv`/`rsDeEnv` keep their full-metadata behavior for round-trip and the non-check loaders; only the check path (`CheckCmd`) switches to `deEnvAnon`.

Consumers updated for the lazy const type: closure walk, IOBuffer ingress (ships `rawBytes` windows directly), shard partitioning, decompile, and the compile/commit env builders.

Measured on mathlib.ixe (same file, same target): 104.5 GB -> 4.5 GB peak RSS, 159 s -> 13.5 s. Nothing in circuit changes: same closure bytes, same in-circuit blake3 verification, same proof.

Builds on the Rust kernel lazy deserialization from #415

### DIFF
--- a/Ix/Cli/CheckCmd.lean
+++ b/Ix/Cli/CheckCmd.lean
@@ -183,7 +183,10 @@ def forEachClaim
   match ixePath with
   | some path =>
     let bytes ← IO.FS.readBinFile path
-    let ixonEnv ← match Ixon.rsDeEnv bytes with
+    -- Anon load: lazy zero-copy constants, binder metadata dropped. The Aiur
+    -- check circuit consumes only anonymous constants, blobs, and per-Defn
+    -- reducibility hints — mirrors the Rust kernel's `get_anon_mmap`.
+    let ixonEnv ← match Ixon.deEnvAnon bytes with
       | .error e =>
         IO.eprintln s!"Failed to deserialize {path}: {e}"; return 1
       | .ok env => pure env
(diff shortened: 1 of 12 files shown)


arthurpaulino (approved):
Nice!
The "perf(aiur)" tripped me up a bit, but the patch looks good.

PR #449, opened 2026-06-17
ci: Add Aiur compile + typecheck benchmark workflow

samuelburnham:
Two main changes:
- Upgrades the `CheckNatAddComm.lean` benchmark to a generic `Typecheck.lean` benchmark, with optional JSON output for CI integration.
- Refactors the bencher.dev `ix compile` benchmark on `main` to also benchmark Aiur execution + proving on the following InitStd and Mathlib constants:
```
- bench: InitStd
  consts: Nat.add_comm Nat.sub_le_of_le_add String.append Array.append_assoc
- bench: Mathlib
  consts: Nat.factorial Nat.Coprime Nat.Prime.two_le
```

Because of Aiur's memory restraints these are all light enough to run easily on the 128GB runner, whereas `Vector.extract_append` hit OOM on execution. These constants should thus be changed to cover more unique workloads as Aiur RAM usage improves.

The prove benchmarks have the following measurements, each with their own thresholds which will fail CI/create an alert when a percentage-based regression from the last 64 runs is detected:
- Constants - 0% threshold, should only change on toolchain bump
- FFTs - 0% threshold
- Execute Time - 10% threshold
- Prove time - 10% threshold
- Peak RAM usage - 10% threshold
- Throughput (constants/second) - 10% threshold

> [!NOTE]
> The `ubuntu-latest` runner builds the `bench-typecheck` binary but it does not have support AVX-512. This means the proving benchmarks will be significantly slower (10-30%), and are mainly meant to test for performance regression than absolute numbers. Later we can upgrade to a more powerful runner.

Successful run:
https://github.com/argumentcomputer/ix/actions/runs/27778973973

After https://github.com/argumentcomputer/tracing-texray/pull/2


arthurpaulino (dismissed):
LGTM!

arthurpaulino (approved):

PR #450, opened 2026-06-19
IxVM kernel: unblock UTF-8 decode/encode proof + Nat-layer FFT cuts

arthurpaulino:
## Headline

Aiur kernel previously **OOM**'d on
`_private.Init.Data.String.Decode.0.ByteArray.utf8DecodeChar?.assemble₄_eq_some_of_toBitVec._proof_1_8`.
This constant is a **prerequisite of `ByteArray.utf8DecodeChar?_utf8EncodeChar_append`**
(part of the UTF-8 round-trip lemma chain).

Now it typechecks at **37.31B FFT** (down from the OOM baseline). The
two-direction lemma `…_utf8EncodeChar_append` itself remains
out-of-reach (still OOMs further along the same dependency tree under
the user's memory cap), but the prerequisite that was the immediate
blocker is unstuck.

The same kernel changes also deliver large FFT cuts on previously-
expensive targets — Tier 1d's structural short-circuit replaces full
delta-whnf cascades wherever it fires:

| Constant | Before | After | Δ |
|---|---|---|---|
| `Vector.append`      | 4.02B | 3.16B | **−21.4%** |
| `Array.append_assoc` | 3.94B | 3.08B | **−21.8%** |

### 1. `a6cd34a` — Tier 1d def-eq short-circuit (the unblock)

Aiur's `k_is_def_eq_core` jumped from Tier 1.5 straight to full delta
WHNF (Tier 2). The cascading Nat.rec / Nat.succ iota expansions then
drove `whnf_const_head` past 1M unique entries before either side
reached a comparable canonical form — OOM. Rust's def-eq settles the
same pair via no-delta whnf + quick structural recursion *before* any
of that fires.

Three minimum-necessary pieces ported. No KStore. No FVar. No Subst /
KernelTypes change.

* **`whnf_nd` family** (`Whnf.lean`, mirror Rust
  `whnf_no_delta_for_def_eq`). Same dispatch tree as `whnf`, but
  `whnf_nd_const_head`'s Defn arm falls through to a stuck
  `apply_spine` instead of delta-unfolding. Iota / proj / quot /
  primitives still fire.

* **`k_infer_only` family** (`Infer.lean`, mirror Rust
  `with_infer_only`). App drops `k_check(a, dom)`; Lam drops
  `k_ensure_sort(ty)`; Let drops val/ty validation. Used at
  `try_proof_irrel`, `is_prop_type`, `try_unit_like` — the def-eq
  tactics that only need the synthesized type.

* **`k_is_def_eq_struct_safe` + Tier 1d wiring** (`DefEq.lean`,
  mirror Rust `quick_def_eq` + post-`try_def_eq_app`). Sort-Sort via
  `level_equal`; Lam-Lam / All-All via recursive `k_is_def_eq` on the
  type and on the body under `Cons(ty_a, types)` (types-cons, NOT
  FVar opening). Inserted between Tier 1c (string lit) and Tier 2
  (full whnf):
  ```
  aw_nd = whnf_nd(a); bw_nd = whnf_nd(b)
  ptr_eq(aw_nd, bw_nd) → 1
  k_is_def_eq_struct_safe(aw_nd, bw_nd) → 1
  try_lazy_delta_app(aw_nd, bw_nd) → 1     (rerun: spine args may have
    reduced past what Tier 1.5's pre-whnf attempt could see)
  ```

Each piece independently validated necessary. The previously-tried
KStore explicit caches, FVar variant + opens, FVar-based binder
opening — all confirmed NOT necessary for this unblock and left out.

3 files changed, +296/−6 lines.

### 2. `36d6c3a` — drop `g_or` from `u64_sub_with_borrow`

`u64_sub_with_borrow` combined two per-byte borrow bits with `g_or`.
The two bits are mutually exclusive: `u_t = 1` ⇒ intermediate `t_i ≥
1` ⇒ subtracting `br_in ∈ {0,1}` cannot underflow ⇒ `u_r = 0`. Field
`+` substitutes for `g_or` directly. Per Aiur cost model `g_or` adds
+1 aux + 1 lookup per call (≈ 5 width); field `+` is free. 7 g_ors ×
2.23M rows.

**UTF-8 `_proof_1_8`: 39.12B → 38.14B (−2.6%)**.

### 3. `9e787da` — drop `g_or` from `klimbs_add_carry` / `klimbs_sub_borrow`

Same mutually-exclusive-carry pattern. Two limb-level borrows /
carries from sequential u64 ops cannot both be 1.

**UTF-8 `_proof_1_8`: 38.14B → 38.07B (−0.18%)**.

### 4. `4e379e7` — hot/cold split `try_nat_dispatch`, extract binop arm

`try_nat_dispatch`'s width 90 was floored by the binop arm (2× whnf
+ 2× try_extract_nat + try_nat_binop_addr + apply_spine), charged on
every Nat.succ / Nat.pred row. Factor binop dispatch into
`try_nat_binop_dispatch`. Main narrows to the max of succ / pred
arms.

**UTF-8 `_proof_1_8`: 38.07B → 37.80B (−0.7%)**.

### 5. `80ce3d2` — hot/cold split `expr_lbr`, extract Let arm

`expr_lbr`'s width was floored by the Let arm (3 recursive expr_lbr
calls + 2 lbr_max + 1 lbr_dec), charged on every row even though Let
is rare. Factor into `expr_lbr_let`.

**Nat.add_comm: 55.63M → 55.50M (−0.2%). UTF-8 `_proof_1_8`: 37.80B
→ 37.62B (−0.5%)**.

### 6. `1f8effd` — hot/cold split `try_extract_nat`, extract App arm

`try_extract_nat`'s width was floored by the App arm (list_lookup +
address_eq + recursive try_extract_nat + klimbs_succ). Factor into
`try_extract_nat_app`. Main narrows to leaf-arm width.

**UTF-8 `_proof_1_8`: 37.62B → 37.31B (−0.8%)**.

### 7. `039e9cf` — re-pin IxVM FFT costs

41 pins in `Tests/Ix/IxVM.lean::kernelCheckEntries` updated. Every
constant got cheaper; none regressed. Largest reductions:
* Vector.append: 4.02B → 3.16B (−21.4%)
* Array.append_assoc: 3.94B → 3.08B (−21.8%)

`lake test -- --ignored ixvm` passes with 0 FFT mismatches.

## Cumulative on UTF-8 `_proof_1_8`

| Commit | FFT |
|---|---|
| baseline (`main`) | OOM |
| `a6cd34a` Tier 1d | 39.12B |
| `36d6c3a` g_or → + in `u64_sub_with_borrow` | 38.14B (−2.6%) |
| `9e787da` g_or → + in `klimbs_add_carry` / `klimbs_sub_borrow` | 38.07B (−0.18%) |
| `4e379e7` hot/cold `try_nat_dispatch` | 37.80B (−0.7%) |
| `80ce3d2` hot/cold `expr_lbr` | 37.62B (−0.5%) |
| `1f8effd` hot/cold `try_extract_nat` | 37.31B (−0.8%) |

**Post-unlock optimization: −4.6%** (39.12B → 37.31B).

## Cost on small targets

* Main baseline `Nat.add_comm`: 56.08M FFT.
* This branch `Nat.add_comm`: 55.50M FFT.

Tier 1d itself adds no overhead on the common case (whnf_nd +
struct_safe + try_lazy_delta_app are themselves Aiur-memoized); the
follow-up optimizations are net wins.

## Test plan

* [x] `lake exe check Nat.add_comm` passes (55.50M FFT).
* [x] `lake exe check Vector.extract_append` passes.
* [x] `lake exe check "_private.…assemble₄_eq_some_of_toBitVec._proof_1_8"`
      passes (37.31B FFT, previously OOM).
* [x] `lake test -- --ignored ixvm` — all 41 FFT pins updated, suite
      passes with 0 mismatches.


Ix/IxVM/Kernel/Infer.lean
```diff
@@ -184,6 +184,108 @@ def infer := ⟦
...
+        let ty_whnf = whnf(ty, types, top, addrs);
+        let pair = ensure_sort_post_whnf(ty_whnf);
+        match pair {
+          (ok, l) =>
+            assert_eq!(ok, 1);
+            l,
+        },
+    }
+  }
+
```
gabriel-barrett, line 325: This is a very nice idea. I believe it's a good idea to explain that it is NOT safe to call `k_infer_only` on a term that was not generated by a well-typed term (i.e. one which `k_infer` was called) though. The reason is that one of our assumptions is that evaluation can only happen after typechecking, and `k_infer_only` will evaluate a type, namely the substitution `B[a/x]` on an application without checking that this substitution is safe (i.e. that `a` has the proper type). The reason this function is okay, though, is that they're being called on terms generated by `whnf` of well typed terms. These are genuinely new terms, but we *know* they're type checked even without checking them.

The only issue with this function is that it doesn't share the cache with `k_infer`. There's a solution we can eventually implement: non-deterministically decide whether to call `k_infer`, `k_infer_only` or don't call them at all. The hint (an enum `None`, `KInfer`, `KInferOnly`) is derived as such:
1. if the term is not unit-like and not a proof, then the hint is `None`
2. if the `k_infer` is a cache hit, then the hint is `KInfer`. otherwise the hint is `KInferOnly`

Then a match statement on the hint will
1. On `None` proceed to deep equality
2. On `KInfer` call `k_infer` and ensure the type is a proof or unit-like
3. On `KInferOnly` call `k_infer_only` and ensure the type is a proof or unit-like

gabriel-barrett (approved):

PR #463, opened 2026-07-01
IxVM: native execution pipeline

arthurpaulino:
Codegen the IxVM Aiur kernel to Rust, run witness/prove Rust-side against a Rust-owned env, and wire the fast path through every `ix check` / `ix prove` entry point.

Wall time on shard 26 of the 64-way `init.ixes` partition: **1029 s → 62 s** (~16× end-to-end).

## What's here

Eight commits, roughly grouped:

**Codegen kernel**
- **Aiur Bytecode → Rust codegen pass.** Walks the compiled IxVM `Bytecode.Toplevel` and emits one `fn aiur_fn_N(...)` per Aiur function into `crates/ix/src/aiur_ixvm.rs` (~4.8 MB, 743 fns). Same `QueryRecord` shape as the interpreter — proof-verification compatible.
- **Value-returning byte ops in the codegen output.** Eliminates the `Vec<G>` scratch buffer per byte op (~245 sites) via per-op helpers (`bytes2_xor_value`, `bytes2_add_value`, etc.) that bump the byte-chip queries and return the gadget output directly. Also dead-folds `unconstrained || false` / `!unconstrained && !false` and uses an unchecked array copy on Call cache hits.

**Rust-owned witness + env**
- **Port `IxVM.ClaimHarness.buildShardCheckEnvWitness` to Rust.** Parallel closure walk (rayon + `DashSet`) + parallel byte→G conversion. Eliminates the per-byte boxing into `Aiur.G` that dominated per-shard wall time.
- **Wire the Rust witness fast path through per-claim check/prove and both benchmarks.** Adds FFIs for `check_addr` / `prove_addr` / `shard_prove`, plus bytes-blob variants for the compiled-Lean-env code path.
- **`EnvHandle`: Rust-owned `ixon::Env` exposed to Lean as an opaque handle.** Collapses six per-call FFIs into four `*_with_env` FFIs + two constructors (`fromIxe` / `fromBytes`). The env is parsed exactly once per CLI invocation; every per-target FFI call reuses the same handle. Prove FFIs return `claim_bytes` so Lean deserializes the wire claim instead of recomputing it.

**CI + ergonomics**
- **`ix codegen --check`.** CI-facing flag that compares the emitted Rust source against the on-disk file and exits 1 on drift. Wired into `lean-test`. Warm ~2 s.
- **`ix check --interp {source|bytecode}`.** Single flag with two modes. `bytecode` bypasses the codegen kernel via the generic Aiur bytecode interpreter — no `ix codegen` + `cargo build` cycle when iterating on `Ix/IxVM/*.lean`. `source` reaches the Aiur source interpreter for its richer per-step diagnostics.

## Coverage

Every `ix check` / `ix prove` shape now routes through the native pipeline:

| Command                              | Path                                           |
| ------------------------------------ | ---------------------------------------------- |
| `ix check NAME --ixe`                | `checkAddrWithEnv` (codegen kernel)            |
| `ix check NAME` (no `--ixe`)         | `checkAddrWithEnv` + `EnvHandle.fromBytes`     |
| `ix check --claim hex --ixe`         | `checkAddrWithEnv` for `check addr none`       |
| `ix check --ixes --shard K`          | `shardCheckWithEnv`                            |
| `ix check --ixes`                    | `shardCheckWithEnv` × all shards, shared env   |
| `ix prove NAME --ixe`                | `proveAddrWithEnv`                             |
| `ix prove --ixes --shard K`          | `shardProveWithEnv`                            |
| `ix prove --ixes`                    | `shardProveWithEnv` × all shards               |
| `Benchmarks/Typecheck` Phase 1 + 2   | shared `EnvHandle` across execute + prove      |

Fallbacks:

- `--interp source` — Aiur source interpreter (richer errors). `Target.leanW` materialised Lean-side.
- `--interp bytecode` — generic bytecode interpreter, skips codegen/rebuild cycle.
- `--claim <hex>` for non-`check addr none` variants (`eval` / `reveal` / `contains` / `checkEnv-with-asm`) — Lean witness builder.

## FFI surface

Six new FFIs replace the ten single-use ones deleted along the way:

```rust
// Constructors
rs_aiur_env_handle_from_ixe(path)     -> LeanExternal<EnvHandle>
rs_aiur_env_handle_from_bytes(bytes)  -> LeanExternal<EnvHandle>

// Check (returns output + ioBuffer + queryCounts)
rs_aiur_toplevel_check_addr_with_env  (toplevel, fun_idx, handle, addr_bytes,  use_bytecode)
rs_aiur_toplevel_shard_check_with_env (toplevel, fun_idx, handle, owned_blob,  use_bytecode)

// Prove (returns claim_bytes + proof + ioBuffer)
rs_aiur_system_prove_addr_with_env    (system, fri, fun_idx, handle, addr_bytes)
rs_aiur_system_shard_prove_with_env   (system, fri, fun_idx, handle, owned_blob)
```

## Measurements

Shard 26 of the 64-way `init.ixes` partition (blake3 / defeq / Nat-heavy):

| Variant                                       |   Wall | Speedup |
| --------------------------------------------- | -----: | ------: |
| Baseline (Lean witness + bytecode interp)     | 1029 s |    1.0× |
| Lean witness + codegen kernel                 | 1015 s |   1.01× |
| Rust witness (serial) + codegen kernel        |  101 s |    10 × |
| Rust witness (parallel) + codegen kernel      |   80 s |   12.9× |
| + value-helper byte ops                       |   64 s |   16.1× |
| + `EnvHandle` (env parsed once)               |   62 s |   16.6× |

FFT cost `107_006_963_281` on every variant — bit-identical `QueryRecord` traces across all backends.

Sanity across three entry points (warm):

| Command                                                          |    Wall | FFT (G)        |
| ---------------------------------------------------------------- | ------: | -------------- |
| `ix check Nat.add_comm` (compiled env, `EnvHandle.fromBytes`)    |   2.4 s | 23_603_449     |
| `ix check --ixe init.ixe Nat.add_comm`                           |   3.1 s | 23_603_449     |
| `ix check Std.Time.Week.Offset.ofMilliseconds` (codegen)         |   9.3 s | 12_430_516_949 |
| `ix check --interp bytecode Std.Time.Week.Offset.ofMilliseconds` |  14.6 s | 12_430_516_949 |
| `ix check --interp source Eq`                                    |   6.3 s | (`Eq: ()`)     |

## New crate

`crates/ix/` (`ix`) — IxVM-specific glue. Depends on `aiur`, `common`, `ixon`.

- `aiur_ixvm.rs` — codegen'd kernel (regenerated by `ix codegen`).
- `aiur_ixvm_runner.rs` — thin `execute_ixvm` wrapper that routes through `execute_generated`.
- `aiur_ixvm_witness.rs` — Rust witness builders (`build_claim_check_witness`, `build_shard_check_env_witness`) with parallel closure walk + byte-to-G conversion.
- `env_handle.rs` — `EnvHandle` wrapping `ixon::Env` with `anon_hints` populated.

## CI

`lean-test` now runs `lake exe ix codegen --check` (exits 1 if `crates/ix/src/aiur_ixvm.rs` doesn't match what the emitter would produce). Prevents merging stale generated kernels that drift from the Bytecode → Rust pass.

## Notes for reviewers

- `crates/ix/src/aiur_ixvm.rs` is generated. Ignore the diff; regeneration is deterministic and CI-gated.
- The `--interp source` path still requires a Lean-side `ClaimWitness`. `forEachClaim` takes a `forceLeanWitness : Bool` and materialises one via `mkWitness` when the source interpreter is selected.
- `EnvHandle` keeps the mmap alive across per-target calls: `Env::get_anon_mmap` returns per-constant `Arc<Mmap>` windows, and the handle owns the Env; Lean's reference-counted `LeanExternal<EnvHandle>` drives the drop.
- The prove-path executor is injected as a closure from `ffi` into `aiur::AiurSystem::prove_ixvm` so `aiur` doesn't need to depend on `ix` (which depends on `aiur`).


samuelburnham (commented):
## Claude review

### 1. Default `ix check --ixes` dropped the coverage check

  `runShardManifestAllNative` (`Ix/Cli/CheckCmd.lean:477`) never calls `shardsCover`; the docstring says "skips the coverage check; trust the user". At the merge base, the all-shards route
  went unconditionally through `runShardCheckAll`, which gates on `shardsCover` first.

  Since the `.ixes` manifest is an untrusted input reconciled against the env by silent intersection (`ownedConstsForBlocks`, `CheckCmd.lean:387`), a stale/truncated/buggy manifest now makes
  `ix check --ixes` exit 0 while some env constants are never checked by any shard. The most likely trigger is the mundane one: edit a definition, rebuild the `.ixe` (content addressing gives
  the edited cone new addresses), forget to re-run `ix shard` — and the constants that go unchecked are exactly the ones just edited. `shardsCover`'s own docstring calls coverage "the whole
  soundness condition for the check case", so exit 0 silently stops meaning "the environment typechecks".

  Both `ixonEnv` and `shards` are already in scope from `loadEnvAndShards`, so the fix is the same one line `runShardCheckAll` uses (`CheckCmd.lean:558`):

  ```lean
  if !(← shardsCover ixonEnv shards) then return 1
  ```

  (Skipping it on the single-shard `--shard K` path is fine/consistent with before; it's the all-shards "full verdict" mode that needs it.)

  ### 2. `ix check --ixe … --claim <hex> --interp source` regressed

  The `claimHex` arm of `forEachClaim` (`CheckCmd.lean:229-234`) maps `.check addr none` to `Target.addr` unconditionally, ignoring `forceLeanWitness` — unlike the two name-based arms right
  below it (`:243`, `:261`), which consult the flag. Under `--interp source`, `runOne` then rejects `.addr` with "`--interp requires a Lean witness; addr/shard targets unreachable here`" — so
  re-checking a persisted `check addr none` claim under the source interpreter always fails (with a message claiming the path is unreachable). At the merge base this combination worked (the
  claimHex arm always built a Lean witness). Fix is to mirror the sibling arms:

  ```lean
  | .check addr none =>
      if forceLeanWitness then
        let witness ← IO.ofExcept <|
          IxVM.ClaimHarness.buildClaimWitness ixonEnv claim trees
        pure (.leanW witness)
      else pure (.addr addr)
  ```

  Ironically only the *common* claim shape breaks — the exotic variants fall through to `.leanW` and interpret fine. And this is exactly the debug path someone reaches for when the fast path
  gives them a failure.

  ### 3. Witness IOBuffer layout is nondeterministic across runs

  `closure_from_set` (`crates/ix/src/aiur_ixvm_witness.rs:119`) collects the closure via `DashSet::new()` (std `RandomState`, fresh seed per process), and that ordering feeds arena `idx`
  assignment in `add_entries_parallel`. Those `idx`/`len` values are committed as auxiliary trace columns (`trace.rs`, `Op::IOGetInfo`), so proof bytes can differ run-to-run and never
  byte-match a Lean-built witness's layout. The `QueryRecord` itself is insulated (all `io_get_info` call sites consume `idx`/`len` locally — so the bit-identical-trace claim in the PR
  description survives), and soundness is unaffected; this is about reproducible proofs and future Lean-vs-Rust differential testing.

  The fix is nearly free: `build_shard_check_env_witness` already computes a sorted `closure_vec` but then passes the *unsorted* set to `add_entries_parallel`; `build_claim_check_witness`
  doesn't sort at all. Sorting both makes the layout canonical.

  ## Worth addressing (this PR or follow-up)

  ### 4. No in-tree parity test

  The whole design rests on "generated kernel ≡ interpreter on the `QueryRecord`", but there's no test or CI step that runs both paths on a fixture and diffs the records (no `#[test]` under
  `crates/ix/`). One record-equality test on a small workload would turn the parity invariant from reviewed-by-hand into checked-by-CI.

crates/ixvm-codegen/Cargo.toml
samuelburnham, line 1: It's a bit unclear what this crate is for based on the name, maybe it could be renamed to `ix-codegen`?

samuelburnham (dismissed):
Stamping to unblock pending CI

samuelburnham (approved):

PR #464, opened 2026-07-01
Aiur kernel: fix Lean.Syntax.rec (shard 53) + `ix name-of`

arthurpaulino:
## `IxVM Aiur kernel: fix Lean.Syntax.rec (shard 53) aux-recursor gen`

`ix check --interp bytecode Lean.Syntax.rec` failed with `assert_eq
mismatch: 0 != 1` on the declared-vs-canonical type equality. Three
interlocking bugs in the Aiur block-flattening / recursor-type builder:

- **Single-pass flat traversal.** `build_flat_block` scanned originals
  once; nested-aux members (`Array Syntax`, `List Syntax`) never had
  their own ctors scanned, so `flat` had 2 motives when Lean's recursor
  declares 3. Replaced with a queue-based fixed point mirroring
  `crates/kernel/src/inductive.rs:build_flat_block:531-599`.

- **`is_rec_field` matched on const idx alone.** For `Lean.Syntax.ident`
  the field `preresolved : List Preresolved` shares the base `List`
  const idx with the block's `List Lean.Syntax` aux, so it got a
  spurious `motive_2 preresolved` IH binder. Match key is now
  `(head_idx, spine-arg prefix ≡ member.spec_params)` — direct members
  carry `spec_params = []` and match on idx alone; auxes require the
  concrete occurrence via new `find_flat_member_match` +
  `spine_prefix_eq` + `kexpr_struct_eq` helpers.

- **`build_all_minors` shrunk `flat` in-flight.** It iterated `flat` and
  passed the shrinking suffix into `build_minor_doms`, so later members'
  ctor-field classification was blind to earlier members. Split into a
  thin wrapper `build_all_minors` + `build_all_minors_walk` that pins
  the caller's full flat while iteration state shrinks.

Every recursor gen call site (`build_minor_at_depth`, `build_minor_doms`,
`build_rule_rhs`, `populate_rules`) now threads the full `flat` alongside
`flat_idxs`.

Ships pinned FFT cost for `Lean.Syntax.rec` in the ixvm test suite and
re-pins every stdlib / primitive / prelude cost that shifted from the
codegen refresh. `List.rec`, `Array.rec`, `Lean.Syntax.rec_1`,
`Lean.Syntax.rec_2` verified to still pass.

## `ix name-of: reverse-lookup a Lean.Name from its content address`

`ix name-of --ixe <path> <hex>`: given a 32-byte hex address, scan the
`.ixe` env's `named` map and print the fully-qualified `Lean.Name` of
the const stored at that address. Exits 1 with an error if no named
const matches (unnamed intermediate const or a blob).

Complements `ix addr-of` (name → address). Primary use: backtrack from
an assertion-failure address logged mid-kernel-execution to the
offending Lean constant.

## Test plan

- [x] `lake test -- --ignored ixvm` passes end-to-end with the new pins.
- [x] `ix check --interp bytecode Lean.Syntax.rec` exits 0.
- [x] `ix check Lean.Syntax.rec` (codegen path, after `ix codegen`) exits 0.
- [x] `ix check --interp bytecode List.rec Array.rec Lean.Syntax.rec_1 Lean.Syntax.rec_2` all exit 0.
- [x] `ix name-of --ixe <ixe> <hex-of-known-named>` prints the expected fully-qualified name; unknown address exits 1.


arthurpaulino:
Superseded by #473

PR #474, opened 2026-07-05
Ixon: .ixe bundle format and diff tool

johnchandlerburnham:
Two related changes

1. the `.ixe` **bundle format** (a single env file that pins one Lean value as a self-contained, verifiable unit — the groundwork for Ixon as a client/server interchange format), plus its first producer CLI, **`ix pack`**
2. **`ix diff`** — a structured, content-address-aware diff of two `.ixe` environments, with **ripple root-causing** (separating intrinsic edits from the content-address ripple) and a memory-lean `--meta`, validated at mathlib scale.

# Part 1 — `.ixe` bundle format

Evolves the format in place (pre-alpha, Tag4(0xE,0) kept, no compat shims). A constant's address already pins its entire dependency DAG (refs tables are content addresses, recursively); this adds the transport half: a distinguished root, an explicit trust boundary for thin bundles, and closure helpers to produce/validate them.

```
[Tag4(0xE,0)] [consts merkle root : 32]
[main : u8 flag + 32? ]              -- NEW: bundle root; readers enforce main ∈ consts
[assumptions : Tag0 count + 32*]     -- NEW: strictly ascending; thin-bundle cut points
§1 blobs    -- NEW: blake3(bytes) == addr verified per entry, every reader
§2 consts   -- unchanged (already verified)
§3 anon_hints  -- moved before metadata; ALWAYS written (derived from Named Def
               -- metadata when the map is empty)
§4 names / §5 named / §6 comms      -- metadata last
```

- **Metadata-last reorder**: `get_anon`/`get_anon_mmap` read §1–§3 and stop — no more parse-and-discard of names/named/comms; the full readers reject trailing bytes.
- **§3 is the canonical hint channel**: read-time hint harvesting is gone. `merkle_root_canonical(assumptions)` reproduces `Claim.assumptions` roots, so claim interop is derivable, not stored.
- **Integrity**: blobs were previously *unverified* in every reader — a swapped blob silently changed a Nat/String literal under an otherwise-valid file. Lean `getEnv` also gains the per-const hash check it was missing.
- **Closure helpers**: `Env::bfs_closure` (3-edge traversal: refs, projection → `Muts` block, block → member/ctor projection addresses), `Env::prune_to_closure(main, assumed)` (minimal closed bundle incl. display metadata), `Env::validate_closed()` (receiver check). `kernel::anon_work::closure_addrs` now delegates to `bfs_closure`.
- **Byte-identical mirrors (bonus fix)**: the Lean and Rust writers were never byte-identical (anonymous-name table entry 0; latent `putIdx` `getD 0` collision). Now locked by strict `serEnv == rsSerEnv` tests — byte-exact at 505,759,024 bytes on the whole-stdlib env.

## `ix pack` — producing bundles

```
lake exe ix pack <env.ixe> <name> [--out <path>] [--assume <name|hex64,...>] [--assume-file <f>] [--verbose]
```

The first production consumer of `prune_to_closure`: resolves `<name>` (displayed form) against the env's `named` table, prunes to the self-contained closure — `main` set, reached cut points recorded in `assumptions`, display metadata carried to fixpoint — re-validates with `validate_closed`, and writes the bundle (default `<name>.ixe`). `--assume` entries (names or 64-hex constant addresses) declare thin-bundle trust boundaries. Euclid's theorem out of the 3.2 GB mathlib env:

```
❯ lake exe ix pack compilemathlib-8850ed93.ixe Nat.exists_infinite_primes --verbose
[rs_pack_env] main 091d37b69086bf163ecfb6c12233e1e741655b8ab09f76979f0bbbdc0c3176b5 (Nat.exists_infinite_primes)
[rs_pack_env] kept 3138/648914 consts, 10181/746073 named, 9504/329107 blobs, 0 assumption(s)
[rs_pack_env] wrote Nat.exists_infinite_primes.ixe (13006199 bytes)
```

3,138 constants (0.5% of mathlib), 13 MB, closed by construction — `main`'s 32 bytes alone pin the value. (`ix shard extract` remains the non-bundle sibling: a general sub-env for the kernel-check pipeline, no `main` root.) Pack v1 reads with the full reader so metadata survives the prune; the 3.2 GB parse dominates its runtime — a leaner selective-metadata pack is future work on the §5 streaming machinery from Part 2.

# Part 2 — `ix diff`

```
lake exe ix diff <old.ixe> <new.ixe> [--anon | --meta] [--verbose]
```

Content-addressing frames the semantics: a name "changed" ⇔ its constant address changed, and value-equality coincides with address-equality — so the interesting work is saying **what** changed, honestly.

```
[diff] old.ixe: 647127 consts, 743832 named, 328281 blobs, 0 comms
[diff] new.ixe: 648914 consts, 746073 named, 329107 blobs, 0 comms
named: 3763 added, 1522 removed, 143528 changed (4516 roots, 139012 rippled)
  + AddEquiv.withBotCongr                33ccd4f845d0…
  ~ AbsoluteValue.exists_partition_int  defn       a382796a804c… → f1cee63a8923…  [type, value]
  ~ RecursiveIn                         iprj→defn  13b38a4e0f38… → 3fe097b1e646…  [kind]
  (139012 rippled rows hidden — address changes fully explained by dependency re-addressing; --verbose lists)
consts: 140144 only in old, 141931 only in new (mutual blocks/projections)
hints changed: 1
```

Exit codes follow GNU diff: 0 = no difference (in the selected mode), 1 = differences, 2 = error. Byte-equal files short-circuit to `identical` without parsing (mmap memcmp; 1.5 s on a 3 GB input).

## Modes

- **`--anon` (default)**: compares only anonymous structure — name→addr changes with per-field classification, consts/blobs set differences, comms, `main`/`assumptions`, and reducibility hints (they live in anon §3 and drive kernel unfolding). Names are join/display keys only.
- **`--meta`**: additionally compares `Named` metadata per component (`meta.info` with kind transitions, `meta.sharing/refs/univs`, `original.added/removed/addr/meta`) — a same-addr metadata-only change is its own report category.

## Field-level classification (Rust engine, `ixon::diff`)

- `exprs_equal` is an iterative lockstep walk that resolves `Share`/`Sort`/`Ref`/ `Str`/`Nat`/`Prj` indices through **each side's own constant tables**: identical expr bytes over different refs tables compare *different*; permuted tables with equal resolution compare *equal*. Pointer-pair memoized (shared subterms don't re-walk), `Share`-cycle-guarded, no recursion (kernel terms nest thousands deep).
- Per-kind field labels (`type`, `value`, `lvls`, `safety`, `rules[i].rhs`, `ctors[j].type`, …). Kind changes report `kind`; an address change with no detected semantic difference reports `encoding` (representation churn) — never an empty list.
- **Projection block descent**: inductives/ctors/recursors live only inside `Muts` blocks behind projections, so "block changed" alone would gut the tool. Same-kind projections with a changed block descend one level: changed member → `block.*` labels; untouched member → `block-siblings` (the block hash moved because a sibling changed — the common, most informative case).

## Ripple root-causing (root vs rippled)

One edited constant re-addresses its whole reverse-dependency cone, so a content-addressed diff drowns the human edits in transitive re-addressings. Every changed row now carries a verdict: after the join, a second pass re-classifies each changed pair under a **quotient** where an (old, new) address pair compares equal when some changed name maps old→new. Rows whose residual labels are all `encoding`/`block-siblings` are **rippled** — fully explained by their dependencies' re-addressing; the rest are **roots** (intrinsic edits). A single-level map is complete because expression comparison only ever consults immediate-dependency addresses — composition across the DAG happens through the per-row verdicts, never `M∘M`.

- `fields` keeps the strict classification; `rippled : Bool` is the orthogonal verdict. The default listing shows roots (plus, under `--meta`, rippled rows carrying metadata edits); `--verbose` lists everything.
- The map is set-valued (name splits tolerated) and built from every changed row, kind changes included — so a `iprj→defn` refactor's dependents still ripple.
- Block descent runs **under the quotient** too: an intrinsically edited ctor's projection is the root (`block.ctor.type`), its untouched siblings ripple, and blocks themselves (unnamed) never enter the map — descent is mandatory or block-internal roots would vanish.
- Documented accepted edges, all fail-safe over-reports: induced re-elaboration (a dependency's universe/arity signature change alters dependents' terms beyond addresses) verdicts root; an intrinsically edited block member with no named projection has no root row of its own; the `block` fallback on assumption-cut bundles verdicts root.
- Alias rows share one cached verdict per (old, new) pair; constants re-parse rather than stash (the deliberate `LazyConstant` no-cache policy keeps memory flat). The pass costs ~one extra classify sweep (+16 s at mathlib scale).

## Memory: `--meta` without the OOM, and an mmap fast path

The first `--meta` design read both envs with the full reader — at mathlib scale that materializes every `ConstantMeta` twice and OOM'd a 124 GiB machine. Both modes now load via the lazy index, and meta mode compares metadata by **streaming both files' §5 named sections in a lockstep merge-join** (§5 is written in ascending name-hash order — exactly `Name`'s `Ord` — in every file): parse one entry per side against its own §4 reverse index, compare, drop. Resident metadata is O(1) instead of everything-at-once.

Raw §5 byte windows are *not* comparable across files — metadata name references are file-relative §4 indices, so identical metadata serializes to different bytes over different name tables. The sweep therefore compares parsed, Address-valued `ConstantMeta` (index-independent), pinned by a §4 index-shift soundness regression test and full-reader parity tests.

The CLI also stopped reading files into Lean `ByteArray`s entirely: `rs_diff_env_files` mmaps both inputs (constant windows stay zero-copy mmap slices backed by the OS page cache) and `rs_ixe_files_equal` does the byte-equal fast path. Anon mode dropped from 54 GiB heap-resident to **8 GiB peak** with a byte-identical report.

## Progress reporting

Large inputs (≥ 100 MB) stream progress to stderr (stdout stays clean for piping): per-env parse timings, then live per-phase events — meta sweep (meta mode), named join, ripple pass:

```
❯ lake exe ix diff compilemathlib-4_29.ixe compilemathlib-8850ed93.ixe
[rs_diff_envs] parsing first env (3211 MB, lazy reader)...
[rs_diff_envs] first env parsed in 12.9s (647127 consts, 743832 named, 328281 blobs)
[rs_diff_envs] parsing second env (3222 MB, lazy reader)...
[rs_diff_envs] second env parsed in 12.1s (648914 consts, 746073 named, 329107 blobs)
[rs_diff_envs] named join: 100000/743832 (19300 changed so far)
...
[rs_diff_envs] named join: 743832/743832 (143528 changed so far)
[rs_diff_envs] ripple pass: 10000/143528 (317 roots so far)
...
[rs_diff_envs] ripple pass: 143528/143528 (4516 roots so far)
[rs_diff_env_files] diff computed in 44.1s
```

## Mathlib-scale validation

Diffed the `v4.29.0` tag env against `8850ed93` (the last mathlib master commit on the v4.29.0 toolchain — 5 days of master; the `Benchmarks/Compile` pin is bumped accordingly): two 3.2 GB envs, 744k named join. **Anon: 65 s wall, 8 GiB peak. `--meta`: 67 s, 8 GiB

```
[diff] compilemathlib-4_29.ixe: 647127 consts, 743832 named, 328281 blobs, 0 comms
[diff] compilemathlib-8850ed93.ixe: 648914 consts, 746073 named, 329107 blobs, 0 comms
named: 3763 added, 1522 removed, 143528 changed (4516 roots, 139012 rippled)
  (synthetic mutual-block names: 627 added, 614 removed — --verbose lists)
  + AddAction.IsPretransitive.of_orbit        2ee60e47dad5…
  + AddEquiv.coe_withBotCongr                 2c55b9612a26…
  + AddEquiv.withBotCongr                     33ccd4f845d0…
...
  - AddCommGroup.ofIsAddCommutative           5ec793909daf…
  - AddCommMonoid.ofIsAddCommutative          c4954ee67492…
...
  ~ AbsoluteValue.exists_partition_int        defn       a382796a804c… → f1cee63a8923…  [type, value]
  ~ AddAction.isClosedMap_quotient            defn       dbba20e21c96… → 4deaf8b7fdc7…  [value]
...
  ~ CategoryTheory.Classifier                 iprj→defn  c632275bad82… → 28096c100344…  [kind]
  ~ CategoryTheory.HasClassifier              iprj→defn  2ba3ede041e1… → 46a6379b87bc…  [kind]
  ~ LinearPMap.mk                             cprj       1df2b7a39e8a… → ae57a82f6164…  [block.ctor.lvls, block.ctor.params, block.ctor.type]
  ~ OmegaCompletePartialOrder.Chain           defn→iprj  b6326c126d8f… → 04e157a9b4cd…  [kind]
  ~ RecursiveIn                               iprj→defn  13b38a4e0f38… → 3fe097b1e646…  [kind]
  ~ RecursiveIn.oracle                        cprj→defn  2796cba80450… → 3d3d403cff32…  [kind]
...
  (139012 rippled rows hidden — address changes fully explained by dependency re-addressing; --verbose lists)
consts: 140144 only in compilemathlib-4_29.ixe, 141931 only in compilemathlib-8850ed93.ixe (mutual blocks/projections; --verbose lists)
blobs: 678 only in compilemathlib-4_29.ixe, 1504 only in compilemathlib-8850ed93.ixe
hints changed: 1
  38bfaaa99e15…  abbrev → regular(2)
[diff] compilemathlib-4_29.ixe ≠ compilemathlib-8850ed93.ixe
```

- 3,763 added / 1,522 removed / **143,528 changed**, of which **4,516 roots and 139,012 rippled** — 96.9% of the changed set is the content-address ripple, explained away by the quotient. Rippled label distribution confirms the shape: 120,941 `[type, value]` + 15,834 `[value]` pure re-addressings.
- The roots are the human-edit surface: all **5 kind changes** (`CategoryTheory.Classifier`/`HasClassifier` `iprj→defn`, `OmegaCompletePartialOrder.Chain` `defn→iprj`, `RecursiveIn`/`.oracle` destructured), all **245 universe-signature changes** (`[lvls, type, value]`), and **8 genuine constructor edits** (e.g. `LinearPMap.mk` gaining `block.ctor.lvls/params/type`). The 613 `block.ctor.type` rows the strict classifier had surfaced turn out to be 8 real edits + 605 induced re-addressings — exactly the noise/signal split the verdict exists for.
- Verdicts are conservative and checkable: zero rippled rows carry a scalar/coordinate label (machine-checked over all 139k), kind/lvls/cidx changes always root, and the report is deterministic across runs.
- `--meta` additionally finds **7,998 names with differing metadata**: 867 metadata-only rows (same address — binder/arena edits like `…._proof_1 [meta.info]`) and 7,131 changed rows with meta labels.
- **Zero `[encoding]` entries** — compilation is deterministic end-to-end.
- **1 hint change** (`abbrev → regular(2)`) on an otherwise-unchanged constant, correctly absent from the named rows (hints live in §3 and don't re-address constants).

## Breaking change

All existing `.ixe` files are stale — regenerate with `ix compile` (readers emit a "pre-bundle-format .ixe; recompile it" hint). FFI `RawEnv` ctor arity is 5→8, `RawEnvLazy` 3→5; `NamedDiff` gains a `rippled : Bool` scalar field.

## Verification

- Rust: full workspace tests green, clippy `--all-features` clean. `ixon` at 247 tests; the diff engine alone carries 34 (refs-shift guard, share-vs-inline, univ-table shift, memo-clearing regression, block descent, lazy/full parity, per-phase progress totals, 14 ripple-verdict tests incl. two-hop chains / name splits / renamed deps / induced re-elaboration, §5-sweep parity vs the full reader, the §4 index-shift soundness regression, mmap-side parity).
- Lean: full `lake test` green; FFI tests incl. self-diff-empty properties over generated envs in both modes (pins marshaling slot order), a ripple-verdict marshaling test (pins the `num_8` slot), file-based-vs-bytes-based diff parity, and six `ix pack` fixtures (closure bundle, removals-only diff subset, name/hex `--assume` cuts, error paths).
- Cross-language: whole-stdlib env byte-exact between writers (505,759,024 bytes); `ix check-rs` 60,601/60,601 (meta) and 51,681/51,681 (anon); `bench-typecheck` executes and proves over the new format.

## Follow-ups

- `prune_to_closure` rescans `env.named` per fixpoint round — worth a reverse index if bulk pruning against mathlib-scale envs becomes common; the same §5 streaming machinery would also let `ix pack` skip the full-reader parse.
- Bundle transport niceties (store sync protocol) remain future work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1802,6 +1802,7 @@ dependencies = [
  "ixon",
  "ixvm-codegen",
  "lean-ffi",
+ "memmap2",
  "mimalloc",
  "multi-stark",
  "n0-error",
--- a/Ix/Cli/DiffCmd.lean
+++ b/Ix/Cli/DiffCmd.lean
@@ -1,0 +1,190 @@
+/-
+  `ix diff <old.ixe> <new.ixe>`: structured diff of two serialized Ixon
+  environments.
+
+  The diff itself is computed in Rust (`rs_diff_envs` → `ixon::diff`):
+  both files are parsed with the full reader and compared on anonymous
+  structure — names serve as join/display keys, constants compare by
+  content address with per-field classification (type/value/lvls/…,
+  `block.*` for projection targets, `"encoding"` when only the
+  representation moved). `--meta` additionally compares named metadata
+  (`ConstantMeta`/`original`).
+
+  Exit codes (GNU diff convention): 0 = no difference found in the
+  selected mode, 1 = differences found, 2 = error.
+-/
+module
+public import Cli
+public import Ix.Address
+public import Ix.Common
+public import Ix.Ixon
+
+public section
+
+namespace Ix.Cli.DiffCmd
+
+private def pad (s : String) (w : Nat) : String :=
+  s.pushn ' ' (w - s.length)
+
+private def shortAddr (verbose : Bool) (a : Address) : String :=
+  if verbose then toString a else ((toString a).take 12).toString ++ "…"
+
+/-- Synthetic mutual-block names embed the block hash as their second
+    component (`Ix.<64-hex>.…`); a changed block churns one such pair
+    per block, so the default display groups them into a count. -/
+private def isSyntheticMuts (s : String) : Bool :=
+  match s.splitOn "." with
+  | "Ix" :: h :: _ =>
+    h.length == 64 && h.all fun c => c.isDigit || ('a' ≤ c && c ≤ 'f')
+  | _ => false
+
+private def printStats (path : String) (s : Ixon.EnvStats) : IO Unit :=
+  IO.println
+    s!"[diff] {path}: {s.consts} consts, {s.named} named, {s.blobs} blobs, {s.comms} comms"
+
+/-- Print up to `cap` addresses (all when `verbose`), one per line. -/
+private def printAddrList
+    (linePrefix : String) (addrs : Array Address) (verbose : Bool) :
+    IO Unit := do
+  let cap := if verbose then addrs.size else min addrs.size 10
+  for a in addrs[0:cap] do
+    IO.println s!"{linePrefix}{shortAddr verbose a}"
+  if addrs.size > cap then
+    IO.println s!"{linePrefix}… and {addrs.size - cap} more"
+
+private def brackets (labels : Array String) : String :=
+  "[" ++ ", ".intercalate labels.toList ++ "]"
+
+private def printNamedSection
+    (d : Ixon.EnvDiff) (wantMeta verbose : Bool) : IO Unit := do
+  let keep (s : String) : Bool := verbose || !isSyntheticMuts s
+  let added := d.namedAdded.filter (keep ·.1)
+  let removed := d.namedRemoved.filter (keep ·.1)
+  let synAdded := d.namedAdded.size - added.size
+  let synRemoved := d.namedRemoved.size - removed.size
+  let metaCount :=
+    if wantMeta then s!", {d.namedMetaOnly.size} metadata-only" else ""
+  IO.println
+    s!"named: {d.namedAdded.size} added, {d.namedRemoved.size} removed, {d.namedChanged.size} changed{metaCount}"
+  if synAdded + synRemoved > 0 then
+    IO.println
+      s!"  (synthetic mutual-block names: {synAdded} added, {synRemoved} removed — --verbose lists)"
+  -- Column width over everything we are about to print.
+  let mut wMax := 0
+  for (n, _) in added do wMax := max wMax n.length
+  for (n, _) in removed do wMax := max wMax n.length
+  for c in d.namedChanged do wMax := max wMax c.name.length
+  for (n, _) in d.namedMetaOnly do wMax := max wMax n.length
+  let w := min wMax 40
+  for (n, addr) in added do
+    IO.println s!"  + {pad n w}  {shortAddr verbose addr}"
+  for (n, addr) in removed do
+    IO.println s!"  - {pad n w}  {shortAddr verbose addr}"
+  for c in d.namedChanged do
+    let kind :=
+      if c.oldKind == c.newKind then c.oldKind
+      else s!"{c.oldKind}→{c.newKind}"
+    IO.println
+      s!"  ~ {pad c.name w}  {pad kind 9}  {shortAddr verbose c.oldAddr} → {shortAddr verbose c.newAddr}  {brackets c.fields}"
+    if wantMeta && !c.metaFields.isEmpty then
+      IO.println s!"      meta: {brackets c.metaFields}"
+  for (n, labels) in d.namedMetaOnly do
+    IO.println s!"  m {pad n w}  {brackets labels}"
+  if d.namedChanged.any (·.fields.contains "encoding") then
+    IO.println
+      "  (encoding = representation changed; no semantic field difference detected)"
+
+def runDiffCmd (p : Cli.Parsed) : IO UInt32 := do
+  let some oldArg := p.positionalArg? "old"
+    | p.printError "error: must specify <old.ixe> <new.ixe>"; return 2
+  let some newArg := p.positionalArg? "new"
+    | p.printError "error: must specify <old.ixe> <new.ixe>"; return 2
+  let oldPath := oldArg.as! String
+  let newPath := newArg.as! String
+  let wantMeta := p.hasFlag "meta"
+  if wantMeta && p.hasFlag "anon" then
+    IO.eprintln "error: --anon and --meta are mutually exclusive"
+    return 2
+  let verbose := p.hasFlag "verbose"
+  let readBin (path : String) : IO (Option ByteArray) :=
+    try pure (some (← IO.FS.readBinFile path))
+    catch e => IO.eprintln s!"error: {path}: {e}"; pure none
+  let some oldBytes ← readBin oldPath | return 2
+  let some newBytes ← readBin newPath | return 2
+  if oldBytes == newBytes then
+    IO.println "identical"
+    return 0
+  let d ← match Ixon.rsDiffEnvs oldBytes newBytes wantMeta with
+    | .error e => IO.eprintln s!"error: {e}"; return 2
+    | .ok d => pure d
+  printStats oldPath d.statsA
+  printStats newPath d.statsB
+  if d.isEmpty then
+    if wantMeta then
+      IO.println "files differ in bytes but no semantic difference found"
+    else
+      IO.println
+        "files differ in bytes but no anonymous-structure difference found (try --meta)"
+    return 0
+  if let some (oldMain, newMain) := d.mainChanged then
+    let fmt : Option Address → String
+      | none => "∅"
+      | some a => shortAddr verbose a
+    IO.println s!"main: {fmt oldMain} → {fmt newMain}"
+  unless d.assumptionsAdded.isEmpty && d.assumptionsRemoved.isEmpty do
+    IO.println
+      s!"assumptions: +{d.assumptionsAdded.size} −{d.assumptionsRemoved.size}"
+    printAddrList "  + " d.assumptionsAdded verbose
+    printAddrList "  - " d.assumptionsRemoved verbose
+  unless d.namedAdded.isEmpty && d.namedRemoved.isEmpty
+      && d.namedChanged.isEmpty && d.namedMetaOnly.isEmpty do
+    printNamedSection d wantMeta verbose
+  unless d.commsAdded.isEmpty && d.commsRemoved.isEmpty
+      && d.commsChanged.isEmpty do
+    IO.println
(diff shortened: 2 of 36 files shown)


samuelburnham (dismissed):

Benchmarks/Compile/lakefile.toml
```diff
@@ -32,4 +32,6 @@ rev = "v4.29.0"
 [[require]]
 name = "mathlib"
 git = "https://github.com/leanprover-community/mathlib4"
-rev = "v4.29.0"
+# Last mathlib master commit on toolchain v4.29.0 (parent of the
+# v4.30.0-rc1 bump); the tag v4.29.0 is 2026-03-30, this is 2026-04-04.
+rev = "8850ed93863d6ce27210a673f9ed235a4e8a2994"
```
samuelburnham: Is this necessary? It makes sense for the benchmark but not sure it should land on main

johnchandlerburnham: yeah that's fair, I'll revert

johnchandlerburnham:
!benchmark compile

samuelburnham:

samuelburnham:

samuelburnham:
!benchmark compile BENCH_ENVS=InitStd,Mathlib

arthurpaulino (approved):

PR #484, opened 2026-07-10
perf: Decrease RAM overhead for `ix compile`

samuelburnham:
# Compile: Mathlib-scale environments in ~17 GB at near-parity wall time

## Problem

`ix compile` materialized several whole-environment-sized structures at once: a
fully-decoded Rust copy of the Lean environment (~25 GB at Mathlib scale),
structured metadata for every named constant (~20× larger than its serialized
form), unbounded per-worker kernel-env caches, and an env-sized output buffer
crossing the FFI. Mathlib peaked at ~103 GB (CI-measured) and FLT at 63–68 GB —
neither compiled on ordinary development hardware.

## Design

Keep every representation compact except the working set actually being read,
always-on, with every configuration producing a **bit-identical `.ixe`**:

- **Lazy Lean-env decode** — constants decode from the Lean-held objects on
  demand, behind a bounded lock-partitioned cache, instead of materializing the
  whole owned Rust copy up front. The Lean heap already holds the environment;
  decode-on-fetch avoids duplicating it.
- **Hot-constant pinning** — the setup scan's reference graph gives exact
  in-degrees before the scheduler starts; the top 16,384 constants go into a
  never-evicted overlay (decode once, lock-free reads), absorbing repeat
  decodes of foundational constants that would otherwise churn any bounded
  cache.
- **Metadata demotion** — accumulator constants and named metadata are held as
  their serialized bytes and decoded on the rare re-read.
- **Fused setup scan** — the ref-graph, groundedness, and inductive-group
  passes share one whole-env decode instead of three.
- **Streamed output** — the `.ixe` writes straight to disk from Rust (`.tmp` +
  atomic rename); no env-sized ByteArray crosses the FFI. The CPU-heavy named
  section encodes in parallel in bounded chunks.
- **Bounded worker caches** — each worker's kernel env clears every 64 blocks
  (a pure cache; clearing is semantics-free).

Hardcoded sizes (cache entries, pin set, segment count) are justified by sweeps
recorded in the commit messages; the segment count scales with hardware
threads.

## Knobs

- `IX_COMPILE_WORKERS=N` — scheduler worker count (default: all cores).
- `IX_COMPILE_DEMOTE=0` — keep structured caches for flows that re-read the
  compiled env in-process.
- `IX_COMPILE_EAGER=1` — decode the whole environment up front on RAM-rich
  machines, trading the largest memory term back for the last few percent of
  wall (FLT: −8% for +13 GiB; Mathlib requires a large-memory machine).

## Results

Local (24 cores, 50 GB hard cap; wall from the compile step):

| env | before | after |
|---|---|---|
| InitStd (105k consts) | 11.2 GiB | **2.7 GiB** / 5.8 s |
| Lean (189k) | — | **4.2 GiB** / 10.5 s |
| FLT (511k) | OOM | **11.6 GiB** / ~53 s |
| Mathlib (737k) | OOM (~103 GiB) | **~17 GiB** / ~78 s |

CI (32-core runner, `!benchmark`): InitStd **1.09× faster** than main at 3.1×
less RAM; Mathlib 5.4× less RAM at +13.7% wall. The residual is repeat-decode
cost that grows with core count (fixed decode CPU over shrinking compute); it
is bounded above by the `IX_COMPILE_EAGER=1` configuration, which the benchmark
tooling can now measure directly.

Byte-identity is verified against the pre-branch binary (`cmp` on
InitStd/Lean; FLT/Mathlib have a pre-existing ±tens-of-bytes named-section
nondeterminism unrelated to this PR).

## Also in this PR

- `!benchmark`: `IX_COMPILE_*` knob passthrough reaches the measured compile
  (cache keys include the knob config, so knob runs publish their own rows).
- `test-ffi` feature builds fixed for the new `Env` API (CI's
  `clippy --all-features` gate).
- sp1 guest adapted to the `Named` metadata accessor.
- The `rust-decompile` suite stays disabled with an accurate note: Rust
  decompile of synthesized `_sparseCasesOn` aux constants fails on main too
  (`missing Ref metadata`) — pre-existing, tracked separately.

## Follow-ups

- Post-merge: `!benchmark compile BENCH_ENVS=Mathlib IX_COMPILE_EAGER=1` to
  bound the many-core residual; if capacity proves to be a worthwhile dial
  there, scale the lazy cache with runner RAM the way segments scale with
  threads.
- File the `_sparseCasesOn` decompile metadata bug upstream.


samuelburnham:
!benchmark compile InitStd Mathlib

samuelburnham:
!benchmark compile BENCH_ENVS=InitStd,Mathlib

samuelburnham:
!benchmark compile BENCH_ENVS=InitStd,Mathlib

samuelburnham:

samuelburnham:

samuelburnham:
!benchmark compile BENCH_ENVS=InitStd,Mathlib

johnchandlerburnham (approved):

PR #485, opened 2026-07-11
IxVM Aiur kernel: drive mathlib shard 0 to green

arthurpaulino:
## Summary

`ix check --ixe mathlib.ixe --ixes mathlib.ixes --shard 0` died with an
opaque `assert_eq mismatch: 0 != 1`. Driving it to green surfaced four
independent correctness divergences between the Aiur kernel and its Rust
mirror, one shard-scale performance pathology, and some missing debugging
infrastructure. All are fixed here: shard 0 (9,357 owned constants of the
647,127-const mathlib env) typechecks end-to-end through the native
codegen kernel, `lake test -- --ignored ixvm` is green (599 assertions),
and the FFT pins land within ±1.1% of `main` (median ratio 0.997) — the
kernel surgery is cost-neutral.

## Kernel fixes

1. **Canonical block sort compares constructors** (`CanonicalCheck.lean`).
   The Indc comparator stopped at the inductive's type, so same-shape
   mutual inductives (`Mathlib.Tactic.Ring`'s `ExBase`/`ExProd`/`ExSum`:
   identical flags, arities and types, differing only in ctor types)
   collapsed into one alpha class and the block was rejected. Now mirrors
   `compare_kindc`'s ctors tail, with a real `KMutCtx` (same-class members
   share one index; ctors mapped at `i + cidx`), address-based external-ref
   comparison, `(fields, rhs)`-only rec-rule comparison, and a provably
   sufficient refinement fuel bound.

2. **whnf-aware index walks** (`Inductive.lean`). `get_result_sort_level`
   and the motive/rec-type index-domain collectors peeled literal Foralls
   only; index binders can hide under definitional wrappers —
   `inductive εClosure (S : Set σ) : Set σ` stores a type ending in
   `Set σ` that only whnf exposes as `σ → Prop`. whnf before every peel,
   mirroring the Rust walks.

3. **Aux synthesis parity + stored-aux validation removal**
   (`Inductive.lean`). Synthesized aux types never instantiated the
   occurrence's universe args (`Type u` vs the stored monomorphized
   `Type 0`) and peeled binders structurally; fixed per
   `canonical_aux_order`'s construction. The stored-aux validation pass
   (`validate_block_auxes` / `is_aux_inductive`) is removed outright:
   stored Muts blocks only ever contain the source originals (Rust seeds
   every stored member `is_aux: false`; auxes are transient detection
   results), and the classification heuristic misclassified originals in
   mixed blocks (`Lean.Compiler.LCNF`'s `Alt`/`FunDecl`/`Cases`/`Code`),
   rejecting legitimate envs.

4. **Canonical aux order** (`Inductive.lean`). Stored recursors bake the
   compiler's canonical aux order into their motive/minor layout, while
   `build_flat_block` discovers auxes in traversal order. The aux suffix
   is now re-sorted by partition refinement over synthetic aux views,
   with sentinel positions (`|top| + ordinal`) standing in for Rust's
   synthetic addresses and **stable** tie handling — an unstable sort
   reverses content-identical auxes whose spec params differ only in
   phantom parameters (`IxVMInd.DedupM`) and breaks rec-type matching.

## Performance

- **Block-membership table**. `derive_block_member_idxs` rescanned all of
  `top` per block with per-block memo keys — |blocks| × |top| query-map
  entries ate 99% of shard-scale CPU. One shared memoized walk now builds
  an rbtree from `addr_key(block_addr)` to `address_eq`-confirmed buckets;
  each query is an O(log N) lookup.

## Tooling

- **`ix name-of`**: new subcommand resolving a content address back to
  its Lean names against a `.ixe` env — potentially many, since
  structurally equivalent constants collapse to the same address (every
  matching entry in the env's `named` table is printed; the one-slot
  `addrToName` index would drop aliases). Unnamed Muts blocks fall back
  to scanning for projections into the block — turns an anonymized
  failing address into an `ix check <name>` fast repro. `ix addr-of`
  (name → address) is unchanged.
- **`dbg!` prints in the codegen kernel**: `emitDebug` was a silent stub;
  it now emits the same `println!` the bytecode interpreter produces, so
  probe builds run at native speed. Production kernels contain no `dbg!`,
  so generated output is unchanged there.
- **`lake exe check` shim dropped**: it had drifted from `ix check`
  (bool `interp` flag vs mode string); one entrypoint, no drift.

## Validation

- Full mathlib shard-0 native run: green, 9,357/9,357 owned consts.
- Bytecode repros green for every constant that failed along the way:
  `Mathlib.Tactic.Ring.ExSum`, `εNFA.εClosure.step`,
  `Lean.Compiler.LCNF.Cases`, plus `Lean.Syntax.rec` /
  `IxVMInd.DedupM.rec` / `IxVMInd.DepthM.rec` for the aux machinery.
- `lake test -- --ignored ixvm`: 599 assertions, 0 failures.
- `verify_bytes_against` was stubbed during the debugging campaign and is
  restored; final pins include blake3 verification and sit within ±1.1%
  of `main`.

## Known follow-ups

- Shards 1–63 have not been run (pure machine time; ~300 GB peak RSS per
  shard, so sequential on a single box).


arthurpaulino:
!benchmark

johnchandlerburnham (approved):

PR #488, opened 2026-07-13
Aiur hash circuits: value-based keccak-256, inlined calls, self-recursive rounds, and a generalized chain-rotate gadget

arthurpaulino:
## Summary

This branch is an optimization arc over the three hash functions proven by
IxVM's Aiur circuits — blake3, sha256 and keccak-256 — targeting the two
costs that matter: **total circuit width** (columns committed per proof)
and **proving throughput** (bytes/s). It also lands the compiler and
benchmarking infrastructure the optimizations needed.

Final state on a 64-core Xeon 6975P-C (512 KiB proven per largest case):

| hash | total width | throughput | dominant circuit |
|---|---|---|---|
| blake3 | 2,510 | ~68 KiB/s | `blake3_g_function` (206 wide) |
| sha256 | 4,043 | ~30 KiB/s | `sha256_round` (342 wide, 64 rows/block) |
| keccak-256 | 5,018 | ~28 KiB/s | `keccak_round` (3,414 wide, 24 rows/perm) |

Highlights along the way: keccak-256 replaced a pointer-heavy
implementation with a value-based one that is **78x cheaper**, and the
closing optimization trio (chain-rotate gadget generalization + the two
self-recursive round restructures) cuts keccak's width by a further
**53%** while raising its throughput **19%**, and cuts blake3's width by
**21%** at unchanged throughput.

## What's in here

**1. Compiler groundwork.**
`@fn(args)` inlined calls: the callee's body is spliced into the caller at
compile time — no separate circuit, no call interface, no output-column
materialization. Hot inner routines (rotations, carry chains, chi/sigma
combinators) are written once and expanded into the circuits that use
them. The pass is total (bottom-up over the inline-call DAG), alpha-renames
callee locals so nested `@`-calls can't capture bindings, and hoists
spliced let-chains out of strict argument positions.

**2. sha256.**
Round-body word ops become byte ops, the message schedule moves into the
block circuit as a memoized shared circuit, ch/maj and the multi-word
additions fuse their carries into single passes with free field adds, and
the 64 rounds run as one narrow called circuit (64 rows/block) instead of
inlining 64x into the block circuit.

**3. keccak-256.**
Rewritten value-based: lanes and the 25-lane state are plain values — no
pointers, no memory traffic; the rate block is absorbed directly from the
byte stream. All 24 rounds share one `keccak_round` circuit.

**4. blake3.**
G-function word ops extracted into `@`-inlined helpers; its three-word
adds fuse into one carry pass.

**5. Measurement + this round of optimizations.**
- A shared bench harness (`lake exe bench-{blake3,sha256,keccak}`) prints
  per-circuit width/height/FFT-cost statistics and the total circuit width
  before the timed runs, over LCG-seeded non-periodic inputs on a
  4K/32K/128K x 1/4-hash grid (512 KiB max per proof — on the throughput
  plateau, fast to iterate).
- The chain-rotate byte-pair gadget generalizes from fixed rotr7/rotr4 to
  any `k` in 1..=7 (`u8_chain_rotr(k, hi, lo)`), making a 64-bit lane
  rotation cost 4 lookups for **any** bit offset.
- `keccak_round` tail-recurses on a round index (iota constants selected
  by a shared-column match), deleting the `keccak_f` driver and its
  24 x 202 call-site columns; rho runs on the chain gadget:
  width 10,604 -> 5,018 (-53%), throughput 23.4 -> 27.9 KiB/s (+19%).
- `blake3_compress_inner_j` likewise self-recurses over its 7 rounds:
  width 3,173 -> 2,510 (-21%), throughput unchanged.
- The statistics cost model's memory-circuit width is corrected to match
  the Rust prover (`size + 7`, not `size + 11`), and kernel-check FFT
  pins are refreshed accordingly.

## Notes for reviewers

- The `Bytecode.Op` change (`u8ChainRotr k` replacing the two fixed ops)
  shifts the FFI constructor tags of `debug`/`u8RangeCheck`/
  `unconstrainedBigUintDivMod` down by one; the Rust decoder is updated in
  the same commit.
- The Bytes2 gadget table grows from 10+16 to 15+31 columns (fixed 65,536
  rows) to serve all seven rotation amounts on channels 13..19.
- The codegen'd IxVM kernel is regenerated in the final commit; until
  then it compiles through legacy fixed-amount aliases.
- A rejected experiment worth knowing about: making sha256's 64 rounds
  self-recursive (rolling 16-word schedule window) halves sha256's width
  to ~1,850 but costs 26% throughput — measured, not modeled — so it was
  dropped under the "never regress throughput" rule. The write-up lives in
  the branch discussion.

## Testing

- `lake test -- --ignored aiur-hashes` — 195 checks: execute, interpret,
  prove and verify for all blake3/sha256/keccak vectors (including rate-
  and block-boundary sizes).
- `lake test -- --ignored aiur` (483 checks) and `lake test -- aiur-cross`
  (340 checks) — op-level vectors incl. the chain-rotate ops, and
  interpreter/bytecode agreement.
- `lake test -- --ignored ixvm` — 599 checks: kernel-check FFT pins and
  codegen/bytecode parity over the regenerated kernel.
- Every commit builds; the two synthetic split points were build-checked
  explicitly for bisectability.

## Follow-up (not in this PR)

- LogUp batching in multi-stark: each lookup currently costs a dedicated
  stage-2 extension column (2 base columns) — ~58% of `keccak_round`'s
  width. Folding fraction pairs into one column would cut every circuit's
  width by up to ~30% and halve stage-2 commitment work.
- Chunked byte streams: the per-byte `take_byte`/`memory[3]` walk is ~5%
  of keccak's cost and ~15% of blake3's.


arthurpaulino:
!benchmark

gabriel-barrett (dismissed):

arthurpaulino:
!benchmark aiur-recursive

arthurpaulino:
!benchmark

samuelburnham:
!benchmark aiur-recursive

PR #492, opened 2026-07-14
Ixon: compact index-keyed .ixe

johnchandlerburnham:
§3 (reducibility hints) and §5 (named metadata) keyed their entries by raw 32-byte addresses that are fully redundant — every §3 key is an address already listed in §2, and every §5 entry repeats a §4 name hash plus a §2 const address. This PR replaces them with Tag0 varint indices, the same move §5 metadata already makes for name references:

- §3: Tag0(index_delta) + Tag0(fused_hint) against §2's ascending-address order (0=Opaque, 1=Abbrev, h+2=Regular(h)) — 2–3 bytes per entry instead of 33–36. Strictly-positive deltas double as the sort/dedup validation.
- §5: name → §4 topo-index, constant → §2 rank (absolute; entry order stays name-hash-sorted for the diff merge-join). original keeps a raw address since a prune cut can reference an assumed constant absent from §2.
- Writers hard-error on hint keys / named entries referencing unstored constants (previously write-OK-read-fail); Lean serEnv is now Except-returning. Readers enforce strictly ascending §2 order, which the indices depend on and which makes the §2 scan double as the sorted merkle key set.
- Lazy paths are untouched: get_anon/get_anon_mmap still stop after §3; parse_lazy_index, NamedMetaCursor, streaming prune, and ix diff --meta stay O(1)-per-entry.

Saves ~56 MB on a mathlib env (~10× shrink of both sections' fixed key costs; proportionally much more on thin ix pack bundles).

Breaking: in-place format change with no version byte (matching the pre-bundle precedent) — existing .ixe files must be recompiled. New readers reject old files with "possibly a pre-compact-ke unaffected (envs are compiled and consumed within a
run).

Rust and Lean writers verified byte-identical (put_file_yte-equality vectors including a new named-entry + hintscase); new tests cover malformed deltas, out-of-range int bytes, and writer key errors.

---

Also in this PR: Compile: keep phantom-spec-param nested auxes distinct in the canonical sort

While testing, lake test -- rust-compile surfaced a pre-existing compiler bug (independent of the format work): sort_aux_by_partition_refinement compared only the instantiated aux types and constructors, so two nested occurrences of one external inductive whose distinguishing spec param is phantom in the external's constructors — the IxVMInd.DedupM fixture shape, Bar2 M Nat vs Bar2 M Bool with Bar2.mk : α → Bar2 α β — produced alpha-identical views and collapsed into one canonical class. compute_aux_perm then (correctly) found no canonical match for the losing source aux and failed the whole block. The sort's doc comment always claimed the nested identity was part of the compared data; it wasn't.

Fix: each aux's sort view carries a trailing synthetic identity-marker ctor holding its aux_to_nested nested-app (block params abstracted to positional bvars). Distinct identities split and order canonically by spec-param content address; genuinely alias-collapsed occurrences (address-equal specs) still dedup; already-distinct auxes are unaffected since the marker is compared last. The kernel's canonical_aux_order mirrors the marker on its synthetic views (ext head + spec_params, pre-rewrite) so both reconstructions stay in lockstep, and compute_aux_perm's no-match error now prints the normalized source specs against every canonical signature.

Coverage: validate-aux now permanently seeds the IxVMInd fixtures (previously these shapes only compiled under the full-env rust-compile runner), plus a unit test pinning split-vs-collapse behavior. lake test -- rust-compile passes end-to-end: all blocks compile, and the full-env re-decompile reports 0 mismatches over 201,979 constants — which also exercises the per-name hints fix at scale.

### DIFF
--- a/Ix/Cli/CheckCmd.lean
+++ b/Ix/Cli/CheckCmd.lean
@@ -293,7 +293,9 @@ def forEachClaim
         let w ← mkWitness addr ixonEnv
         runOne claim none (.leanW w) label
       else
-        let envBytes := Ixon.serEnv ixonEnv
+        let envBytes ← match Ixon.serEnv ixonEnv with
+          | .error e => throw (IO.userError s!"serEnv failed for {label}: {e}")
+          | .ok b => pure b
         let envHandle ← match Aiur.EnvHandle.fromBytes envBytes with
           | .error e => throw (IO.userError s!"EnvHandle.fromBytes failed for {label}: {e}")
           | .ok h => pure h
--- a/Ix/CompileM.lean
+++ b/Ix/CompileM.lean
@@ -1622,10 +1622,16 @@ def compileEnv (env : Ix.Environment) (blocks : Ix.CondensedBlocks) (dbg : Bool 
   -- Merge name string blobs into the main blobs map
   let allBlobs := nameBlobs.fold (fun m k v => m.insert k v) compileEnv.blobs
 
-  -- Resolve per-name hints to each name's registered constant address
-  -- (the projection address for mutual-block members — exactly the
-  -- address the kernel looks hints up under). Alias collisions merge
-  -- order-independently, matching Rust `CompileState::finalize_hints`.
+  -- Resolve per-name hints into both channels (matching Rust
+  -- `CompileState::finalize_hints`): the EXACT value onto each Named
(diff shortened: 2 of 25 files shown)


johnchandlerburnham:
!benchmark

samuelburnham:

johnchandlerburnham:

samuelburnham (approved):

PR #497, opened 2026-07-17
Blake3 circuit optimizations + Aiur inlined function calls

arthurpaulino:
Distills the blake3 work from `ap/aiur-hash-width` onto `main`: the `@`-inlining
compiler feature plus three blake3 circuit optimizations. Excludes the
chain-rotr gadget generalization (blake3 only needs the existing fixed
`u8_chain_rotr4`/`u8_chain_rotr7` ops), the keccak/sha256 work, and the bench
harness changes.

## Commits

1. **Aiur: inlined function calls (`@fn(args)`)**

   Replaces `Source.Term.app`'s `unconstrained : Bool` with a `CallMode`
   (normal | unconstrained | inlined). `@fn(args)` marks an inlined call:
   `Toplevel.inlineCalls`, run first in `compile`, splices the callee's body
   into the caller. An inlined call compiles to no separate circuit and no
   call interface — the callee's work joins the caller's straight-line
   circuit. The mode only exists in the Source stage; evaluation is unchanged.

2. **Aiur blake3: extract G-function word-ops into @-inlined helpers**

   `blake3_g_function` was ~100 lines of inlined byte arithmetic. Rewritten
   as its eight-step dataflow over `@`-inlined word ops: ByteStream's
   existing `u32_add` / `u32_xor`, plus four right-rotations
   `u32_rotr16`/`u32_rotr8` (byte moves) and `u32_rotr12`/`u32_rotr7`
   (byte move + chained gadget), added to ByteStream next to the other
   unsigned-int ops. Still one circuit, byte-identical work; FFT unchanged.

3. **blake3: fuse the G-function's 3-word adds into one carry pass**

   `u32_add3` (ByteStream) accumulates each byte column with chained `u8_add`
   and folds the 0/1 carry bits with a free field add: 11 u8_adds vs 14 for
   two chained adds. G-function lookups 63 → 57, width 224 → 206.

4. **blake3: self-recursive compress rounds**

   `blake3_compress_inner_j` recurses on the round index, permuting the
   message half between rounds and returning after round 6, so
   `blake3_compress` pays one 128-output call site instead of seven
   (1118 → 338 columns). Includes the regenerated Aiur kernel
   (`lake exe ix codegen`) and the 56 re-pinned kernel-check FFT costs.

## Results

`lake exe ix check Nat.add_comm` (blake3 drives every kernel-check's hashing):

| | main | this branch | delta |
|---|---|---|---|
| Total width | 35,705 | 34,964 | −2.1% |
| Total FFT cost | 53,748,034 | 52,451,949 | −2.4% |

All 56 pinned kernel-check FFT costs re-measured from a full
`lake test -- --ignored ixvm` run (suite green, codegen/bytecode parity
included).


arthurpaulino:
!benchmark aiur-recursive

samuelburnham:
!benchmark

PR #502, opened 2026-07-18
perf: IxVM Aiur kernel FFT-cost reduction (−19.4%)

arthurpaulino:
Six measured optimizations to the Aiur kernel and its Ixon ingress,
reducing total FFT cost on the benchmark constant
`Std.Tactic.BVDecide.BVExpr.bitblast.goCache_Inv_of_Inv._mutual`
from **260.46B to 209.92B (−19.4%)**. Every commit was gated on a full
`ix check` run of that constant (build + ingress + checking its entire
5,815-constant dependency closure) and regenerates
`crates/ixvm-codegen/src/aiur_ixvm.rs`; the check passes at every
commit. No accept/reject semantics change on well-formed input.

## Commits

| Commit | Change | Total FFT cost |
|---|---|---|
| `b031fe0` | Lazy valid-cut chain for `ctx_reachable` | 260.46B → 230.48B (−11.5%) |
| `75cd384` | Cold-extract wide arms of the substitution walks | → 224.38B (−2.65%) |
| `ac087c3` | Inline the byte-stream uncons in hot Ixon readers | → 220.68B (−1.65%) |
| `040889f` | Canonicalize def-eq pairs by pointer order | → 218.87B (−0.82%) |
| `8fce786` | Cold-extract `k_infer_core` arms, tighten `inst_many` bvar test | → 214.82B (−1.85%) |
| `7eadf19` | Return `Expr` by pointer from the Ixon reader family | → 209.92B (−2.28%) |

Highlights per commit:

- **Valid-cut chain**: replaces the restart-based reachable-context
  fixpoint (8.5% of total cost, zero memo hits) with a cut chain
  materialized lazily in the memo table — each link keys on one list
  pointer, so a context suffix's chain is computed once and shared by
  every query and binder depth. The structural analogue of the Rust
  kernel's `ctx_addr_cache`.
- **Arm extractions** (two commits): rare match arms carrying multiple
  call sites (Proj/Let/BVar arms of the substitution walks and
  `k_infer_core`) were widening every row of their hot circuits;
  extracting them to their own circuits charges the width only on the
  rows that take the arm (`k_infer_core` width 91 → 52 with its Proj
  arm executing 65K of 2.64M rows).
- **Uncons inlining**: `read_byte` was one full dispatch row per input
  byte whose only work was a memory lookup the caller can absorb for
  one column; 10.6M rows collapse to the cold once-per-constant
  parsers. Truncated input now fails the prover instead of
  zero-padding; well-formed Ixon never over-reads.
- **Def-eq canonicalization**: def-eq is symmetric, so pairs are sorted
  by pointer at the core memo boundary (once per unique pair, not per
  call). The measured win is mostly orientation: the eta/structural
  recursion issues its substitution traffic in one direction, deduping
  ~1M instantiation rows across mirrored queries.
- **Pointer returns**: the Ixon expression readers returned the `Expr`
  union by value, taxing every row with the widest variant's columns;
  returning `&Expr` narrows `get_expr` 75 → 48 and
  `get_app_telescope` 78 → 41 with an unchanged memory-table footprint.

## Cost-model notes (baked into the changes, useful for review)

Measured on this workload and verified to two decimals:

- **FFT cost = width × height × log2(height)** per circuit.
- **Width ≈ base + ~4 columns per call site, summed over all match
  arms** — mutually exclusive arms do not share columns. This is why
  hot/cold arm extraction works and why inlining branching dispatch
  into hot arms does not.
- A memo hit is ~free and truncates the entire call subtree; `load`s
  of resident records add no rows. Unique argument tuples, not call
  counts, drive cost.

## Explored and rejected (measured, then reverted)

- **N-ary application spines**: full migration built and semantically
  validated (the bitblast check passes), but measured +4.7% to +7.7%:
  a closed arg in a binary spine costs a free entry memo hit, while an
  arg-list walker pays a full-width row per cell it passes. Binary
  chains + memoized per-edge dispatch are effectively optimal under
  this cost model.
- **Inline per-child dispatch in substitution walks**: +4.2% — walk
  widths tripled (27 call sites) for entry rows the memo table was
  already amortizing.
- **`collect_spine` snoc → cons+reverse**: −0.13%, below the
  keep-it bar; the accumulator-threaded reverse shares nothing.
- **Bidirectional telescope checking**: rejected at estimate (~0.1-0.2%
  ceiling), and skipping per-binder domain sort-inference would accept
  values the current kernel rejects.

## Semantics notes

- The valid-cut chain relies on frame well-formedness (a stored type's
  loose-bvar range never exceeds the depth below it) — maintained by
  the kernel and asserted by the Rust mirror; malformed contexts now
  fail the prover instead of clamping.
- Def-eq pair ordering changes which side is walked first, not
  verdicts: every tier treats the sides symmetrically or tries both
  directions.
- `expr_inst_many_bvar`'s window test drops a branch that is dead under
  the walk invariant (`lbr(BVar i) > depth ⇒ i ≥ depth`); a violating
  offset wraps in the field and fails the u32 range decomposition, so
  there is no silent path.


arthurpaulino:
!benchmark

johnchandlerburnham (approved):

PR #503, opened 2026-07-20
Recursive verification of Ix proofs at kernel scale

arthurpaulino:
This branch makes the MultiStark verifier runnable *inside* IxVM — as an
Aiur program whose execution can itself be proven — and then optimizes
recursive verification from "never finishes" to a **complete, sound,
end-to-end run**: a kernel-scale proof of `Nat.add_comm` is proven,
recursively verified in-circuit, and the verifier's execution proven in
**4:00 wall clock at 315.9 GiB peak RSS** on a 495 GiB host, producing a
**12.3 MB outer proof that natively verifies in 94 ms**.

## What's here

**Foundation** (`67dcde3`): `ix codegen` generates Rust for the
Ix/MultiStark verifier circuits, and the IOBuffer (proof bytes +
verifying key) is populated natively in Rust, so the recursive verifier
runs as compiled code over IO advice instead of interpreted source.

**Optimization campaign**, each step gated on the reference-vector
self-tests, the e2e accept/tamper/codegen-parity suite, the kernel FFT
pins, and toy + kernel-scale canaries:

- **Native Goldilocks arithmetic** (`5335a17`, execute 5.3×): the Aiur
  native field *is* Goldilocks, so field ops become native ops with
  byte decomposition kept only where the protocol needs canonical bytes.
- **Indexed proof reads with unconstrained byte fetches** (`d674f97`):
  proof-stream bytes arrive by indexed `io_read` instead of walking a
  per-byte list; fetching is unconstrained, with soundness recovered by
  the challenger binding.
- **Direct blake3 for Merkle 2-to-1 compressions** (`b96eb37`) and
  **lane-granularity leaf hashing** (`f6a62b1`): MMCS hashing works on
  64-byte blocks and u64 lanes instead of byte streams.
- **Linear challenger observation** (`f89658f`): observation buffers are
  prepend-built and concatenated once — the quadratic snoc-per-item
  transcript growth was the original "never finishes" cause.
- **Segmented, hugepage-backed, hash-caching QueryMap** (`9161a9f`):
  mmap'd fixed segments with `MADV_HUGEPAGE` before first touch and
  stored per-entry hashes, removing the memoization table as the
  system-level bottleneck.
- **Toplevel pruning** (`8aca57e`): production VMs keep only the
  reachable closure of their entrypoints.
- **Verifying-key ingestion by IO slices and indexed reads**
  (`e7c83f1`, execute −12%): the vk is hashed straight from the IO
  arena in 64-byte blocks and parsed via indexed reads.
- **Native extension-field representation** (`7d12c1f`, execute −40%,
  FFT −27%): `Ext` becomes a native `(G, G)` pair inside the arithmetic
  core; canonical bytes survive only at true protocol boundaries
  (ingest, challenger observation, digest comparison).
- **Concat-free leaf hashing** (`555af77`): the lane hasher walks
  rows-of-lanes directly instead of materializing a concatenated list
  per FRI query.
- **Sparse kernel proofs** (`2c6267c`, inner proof −33%): adopts
  multi-stark's per-proof circuit activation — circuits the checked
  claim never touches emit empty traces and are neither committed,
  opened, accumulated, nor constraint-evaluated. The activation bitmap
  is bound into the Fiat-Shamir transcript before any challenge; the vk
  stays canonical so the recursion statement's `system_digest` binding
  is unchanged.
- **Trivial-helper inlining** (`25479ff`): relic constant helpers from
  the byte-array era become literals; verifier width −172 columns.
- **Split-streams vk encoding** (`ccaaa05`, vk 7.8× smaller): the old
  format was 96% fixed-width padding around 3% payload. Each field
  class now lives in a per-circuit fixed-width segment (1-byte packed
  tags, u16 indices, size-classed constants, u32 metadata; derivable
  `degree_multiple` dropped), so every in-circuit read stays a
  static-size `io_read`. This removed the dominant blake3 cost — the
  vk digest binding — and is what lets the outer prove fit in memory.

## Measurements (kernel-scale `Nat.add_comm`, q=100, logBlowup=2)

Recursive execution of the in-circuit verifier:

| Milestone | Execute | FFT cost | Inner proof | vk |
|---|---|---|---|---|
| Campaign start | never finishes | — | 35.02 MB | 8.38 MB |
| Linear challenger + native G + IO reads | 59.7 s | 229.5B | 35.02 MB | 8.38 MB |
| vk ingestion overhaul | 52.7 s | 222.0B | 35.02 MB | 8.38 MB |
| Native extension field | 31.5 s | 162.0B | 35.02 MB | 8.38 MB |
| Concat-free leaf hashing | 29.0 s | 158.6B | 35.02 MB | 8.38 MB |
| Sparse proofs | 26.0 s | 149.9B | 23.29 MB | 8.38 MB |
| **Split-streams vk (final)** | **18.9 s** | **99.8B** | 23.29 MB | **1.08 MB** |

End-to-end sound recursion (first ever completion; previously the outer
prove OOM'd the 495 GiB host):

| Phase | Result |
|---|---|
| Inner prove | 2.0 s, 23.29 MB proof |
| Recursive execute (in-circuit verify) | 19.8 s |
| Outer prove (over that execution) | 216.9 s |
| Outer proof | 12.3 MB, native verify 94 ms |
| **Total** | **4:00.42 wall, 315.9 GiB peak RSS** |

The remaining recursive-execute cost is dominated by MMCS/leaf/
challenger blake3 (irreducible under the blake3-PCS choice) and the FRI
query loop; the vk — formerly ~35% of FFT — is now ~7%. Merkle-izing
the vk was measured and deliberately rejected: at the new sizes, auth
paths cost as much as the inactive bytes they'd skip for kernel-scale
claims, and proof folding (the intended future) converges to full
activation where flat hashing is optimal.

## Testing

- `lake test -- --ignored multi-stark` — reference-vector self-tests +
  differential hash tests.
- `lake test -- --ignored recursive-verifier` — e2e accept, tamper
  rejections, codegen↔interpreter parity.
- `lake test` / `lake test -- --ignored ixvm` — kernel suites, FFT pins.
- `cargo test -p aiur vk_codec` — codec round-trip fixpoint,
  degree-recomputation equality, malformed-input rejection.
- `lake exe bench-recursive-verifier --queries 100 --execute-only` — toy
  canary (0.68 s / 3.96B FFT, `RAYON_NUM_THREADS=1`).
- `lake exe ix codegen --check` — generated Rust in sync.
- Measurement tooling: `IX_DUMP_RECURSION_IO=<dir>` dumps the
  proof/vk/claims advice blobs for offline format analysis.


arthurpaulino:
!benchmark aiur-recursive

arthurpaulino:
!benchmark aiur-recursive

johnchandlerburnham (approved):

PR #508, opened 2026-07-22
aiur: adopt multi-stark flat lookup witness (LookupValues)

arthurpaulino:
multi-stark now stores the lookup witness flat per circuit instead of nested per-row vectors, removing one heap allocation per lookup per row and freeing the witness before the commit/quotient/FRI stages. Convert each circuit's nested lookups via `LookupValues::from_rows` inside the witness map so the nested form is a per-circuit transient.

Bumps multi-stark to b3a62f7 (perf/lookup-ram).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2076,7 +2076,7 @@ dependencies = [
 [[package]]
 name = "multi-stark"
 version = "0.1.0"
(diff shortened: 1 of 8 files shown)


arthurpaulino:
!benchmark

arthurpaulino:
!benchmark

arthurpaulino:
!benchmark

arthurpaulino:
!benchmark aiur-recursive

johnchandlerburnham (approved):

PR #511, opened 2026-07-23
ci: Bench fixes

samuelburnham:
- Simplifies `bench-main.yml` to derive all parameters from `Ix/Cli/BenchCmd.lean` rather than hardcoded across CI files. This simplifies the matrix job structure and maintenance going forward.
- Sets all of `InitStd,Lean,FLT,Mathlib` as the default `BENCH_ENVS` for `!benchmark compile/decompile`
- Adds a `BENCH_CONSTS` env var to allow overriding the default inputs with any input from the above four envs, which are already compiled to .ixe files in the prior job
- Fixes the recursive verifier plot titles on bencher.dev and makes the script declarative by deleting any old/untracked plots and replacing them with exactly the printed (pass --dry-run to see the plan)

### DIFF
--- a/.github/workflows/bench-main.yml
+++ b/.github/workflows/bench-main.yml
@@ -2,34 +2,24 @@ name: Benchmark main
 
 # Benchmarks tracked on Bencher on every push to main, all reusing the one
 # compiled `.ixe` so the compiler runs once. Every measurement is a single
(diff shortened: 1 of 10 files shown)


arthurpaulino (approved):
Approving to unblock broken benchmarking

PR #516, opened 2026-07-24
multi-stark: adopt flat quotient commitment (rev 678f8e1)

gabriel-barrett:
Bump multi-stark to 678f8e1c96b3d67eaddaa9aa8d03cd7695f97081, which commits each circuit's quotient coefficient slices as one wide matrix on the trace domain (opened once at ζ) instead of one matrix per slice on the split sub-cosets.

Port the recursive verifier to the new proof shape:
- OOD: quotient(ζ) is now the power series Σᵢ ζ^(i·n)·cᵢ(ζ) over the slice values reconstructed from the single opened row, replacing the sub-domain Lagrange recombination (zps products over vanishing polynomials). Adds the check_shapes width assert (qd slices per row) and the one-matrix-per-circuit round-length check in verify().
- PCS: the quotient FRI batch is one matrix per circuit at the same per-circuit heights as the stages; the per-chunk open loop and the chunk-expanded height list are gone, and the verifying-key circuit list no longer threads through the PCS phase.
- Delete the dead zps machinery (g_pow, vanish_chunk, zps_prod, quotient_sum, log_qd_of, open_q_chunks, heights_quotient, repeat_g).

Regenerate crates/ixvm-codegen/src/aiur_multi_stark.rs (215 aiur fns).

Verified: multi-stark self-tests and the recursive-verifier end-to-end suite pass (honest proof accepted, codegen/interpreter parity holds, tampered proof and tampered claim rejected).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2076,7 +2076,7 @@ dependencies = [
 [[package]]
 name = "multi-stark"
 version = "0.1.0"
-source = "git+https://github.com/argumentcomputer/multi-stark.git?rev=c178aaedaaea003331f0992d385266068cee32c8#c178aaedaaea003331f0992d385266068cee32c8"
+source = "git+https://github.com/argumentcomputer/multi-stark.git?rev=40092c160f51fdc2330b9e388873ece3fa4c12c2#40092c160f51fdc2330b9e388873ece3fa4c12c2"
 dependencies = [
  "bincode",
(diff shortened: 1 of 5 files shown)


gabriel-barrett:
!benchmark aiur fresh

gabriel-barrett:

arthurpaulino (approved):

PR #524, opened 2026-07-30
feat(aiur): Measured-ingress sharding

samuelburnham:
# Measured-ingress sharding for Aiur checking and proving

## What this does

Until now, each shard shipped the **full transitive dependency closure** of its constants: everything reachable by following references, and the references of those, and so on. That closure is a huge over-approximation, because *reachable* is not *read*. The kernel always needs a referenced constant's **type**, but it only opens a constant's **body** when definitional equality actually forces an unfold — and most of a closure is only reachable *through* bodies that are never opened. Example: checking a theorem references lemmas; each lemma's proof mentions thousands more constants; the kernel checks the theorem without ever looking inside those proofs — it only needs the lemmas' statements. In practice the kernel reads a small portion of the closure, but the closure was what every shard carried.

This PR improves performance by making `ix profile` run the out-of-circuit Rust kernel over the whole env, and for each typechecked constant it records the set of constants the kernel actually consulted during that check — read a type, or unfolded a body. This "touch graph" is stored in the `.ixprof` file for use by the Aiur shard planner. The planner then gives each shard the full bytes of exactly the constants its checks consulted, plus small type-only stubs for everything referenced but never consulted. Soundness never rests on the measurement: if a shard turns out to need something the recording missed, its check fails closed, naming the missing constant — and the repair driver ships that one constant and retries.

## Performance win

Same 250 GiB per-shard budget, before → after:

| env | shards | smallest possible budget | expected total prove time* |
|---|---|---|---|
| Init | 699 → **60** | 174.8 → **25.6 GiB** | **~1.5 h** |
| InitStd | 11,749 → **118** | 482.9 (infeasible) → **88.6 GiB** | **~3 h** |
| Lean | 56,293 → **174** | 538.2 (infeasible) → **88.6 GiB** | **~4 h** |
| Mathlib | 227,150 → **2,678** | 3,829.7 → 365.8 GiB (one oversized block remains) | ~2 days† |

\* sequential on one 256 GB box: the packer's per-shard prove predictions, calibrated by the ~1.5× ratio observed on measured shard proves.

† Mathlib needs a >365 GiB box for its one oversized block anyway, and the batch prover parallelizes across shards with `--jobs` given the RAM.

All 60 Init shards check green and spot-checked shards prove and verify on a 256 GB box (heaviest measured 189 GiB, ~2 min per shard). Cost: the kernel hooks add **+0.32% mean / +0.47% max** FFT to ordinary per-constant checking.

## Major pieces

- **Touch graph recording** (`ix profile`, `.ixprof` v4) — zero cost outside profiling; the guest kernels are unaffected. The summary now prints per-metric leaderboards (`--top N`: heaviest blocks by heartbeats, substitutions, ingress bytes, and predicted cost, priced by `--backend aiur` / `zisk` / `all`), and `ix profile sweep` predicts the full-closure cost of every named constant in the env.
- **Repair driver** (`ix check --repair`) — the Aiur kernel replays slightly differently than the recording kernel, so a few shards need more than they measured. The kernel now *names* the constant it is missing, and the driver escalates exactly that block. Init converges 60/60 green with no RAM growth.
- **Batch prove** (`ix prove --ixe E.ixe --ixes M.ixes`) — one setup shared across all shards, heaviest first, resumable: a proof is recorded (in `~/.ix/cache/shard-proofs/`, keyed by claim digest) only after it verifies and binds to the shard's reconstructed claim.
- **Address-only stubs** — at prove time, stubs the execution never consults ship as bare addresses (no bytes, no hashing): −5–10% prove RAM.
- **`aiur-shard` benchmark backend** — tracks predicted shards, ingress bytes, and real execution time/RAM of the heaviest shards per env in bencher; Init added as a registry env. `!benchmark aiur-shard` works per-PR.
- **End-to-end regression test** — env → profile → pack → check → batch prove → composed verdict → resume, in the gated `ixvm` suite (~15 s).

## Reproduce

```sh
lake build ix

# 1. compile an env and record the touch graph
ix compile Benchmarks/Compile/CompileInit.lean --out Init.ixe
ix profile Init.ixe --backend aiur                      # → Init.ixprof

# 2. pack + check every shard, escalating divergent ones until all green.
#    With --repair, --ixes is the OUTPUT path: this command creates
#    Init.ixes from the profile and repacks it in place as it escalates
#    (~15 min for Init).
ix check --ixe Init.ixe --ixes Init.ixes --repair \
  --ixprof Init.ixprof --max-ram 250 --jobs 8

# 3. prove all shards (resumable; re-run the same command after a crash)
ix prove --ixe Init.ixe --ixes Init.ixes

# 4. verify any printed proof address
ix verify <proof-hex>

# benchmark row (planner metrics + heaviest-shard executions)
ix bench run --backend aiur-shard --env Init
```

The E2E test runs in `lake test -- ixvm --ignored` (also in PR CI).


samuelburnham:
!benchmark

samuelburnham:
!benchmark aiur-shard

samuelburnham:
Superseded by #601

PR #525, opened 2026-07-30
multi-stark chained logUp: adopt in the recursive verifier

gabriel-barrett:
# Chained-accumulator logUp: adopt in the recursive verifier

Adopts the multi-stark chained-accumulator logUp
([argumentcomputer/multi-stark#`logup-optimization`](https://github.com/argumentcomputer/multi-stark/tree/logup-optimization),
rev `2ecdd36`). Stage 2 no longer commits message inverses: instead of
`1 + L` extension columns per circuit (running accumulator + one
inverse per lookup), it commits **`max(L, 1)` partial accumulators**,
where slot `j` holds the running sum *entering* lookup `j`'s step on
that row.

## The constraint system

One constraint per lookup — down from `L + 3`:

```text
interior (j < L−1):  m_j · (acc_{j+1} − acc_j) − mult_j = 0
wrap     (j = L−1):  m_j · (acc_0′ − acc_j + is_last_row·Δ) − mult_j = 0
```

with `m_j = β + fingerprint(γ, args_j)`, `acc_0′` the next row's first
slot, and `Δ = acc_final − acc_initial` (the existing publics).

The wrap step needs no transition selector: on the trace subgroup
"next row" is rotation by the generator, so the accumulator chain is a
**cycle**. Telescoping the steps around that cycle cancels every
accumulator term and forces the circuit's multiset sum to zero; the
`is_last_row·Δ` term perturbs exactly one link, converting that into
"the multiset sum equals the public accumulator difference". The
first-row, transition, and last-row constraints are all gone, and the
committed accumulator is gauge-free (only differences are constrained;
the prover starts it at zero). Degrees are uniform
`max(deg(args), deg(mult)) + 1` — never above the old profile.

**Selector normalization.** The wrap constraint consumes `is_last_row`
*additively*, so its scale is load-bearing: Plonky3's selector is the
unnormalized Lagrange numerator, with value `n·g` at the last row
(pinned in multi-stark by a test against the textbook Lagrange
product). Prover and verifier now normalize `is_first_row` by `1/n`
and `is_last_row` by `1/(n·g)` protocol-wide; vanishing-form user
constraints are unaffected by the rescale.

## Changes in this PR

- **Rust: none.** The vk codec derives `stage2_width` and
  `constraint_count` from multi-stark's own functions (nothing
  derivable is serialized), so the wire format is unchanged and the
  rev bump alone makes `cargo test -p aiur` pass.
- **`Ix/MultiStark/Verifier.lean`**: `logup_steps_fold` replaces the
  message-inverse fold — one chained step per lookup, the wrap
  detected by index against the lookup count, the injection built from
  the publics; the three boundary-constraint folds are deleted.
  `trace_selectors` returns normalized selectors (new `pow2` helper,
  scales `1/n` and `1/(n·g)`).
- **`Ix/MultiStark/SystemDeserialize.lean`**: derived Fiat–Shamir
  shape values become `zeros + max(L,1)·D` and `max(L,1)·D`.
- `crates/ixvm-codegen/src/aiur_multi_stark.rs` regenerated
  (`lake exe ix codegen`).


gabriel-barrett:
!benchmark aiur fresh

gabriel-barrett:

gabriel-barrett:

arthurpaulino (approved):

PR #526, opened 2026-07-30
logUp grouping

gabriel-barrett:
# Adopt multi-stark logUp grouping; raw lookup arguments on branchless functions

Adopts multi-stark's circuit-local logUp message grouping (branch `logup-grouping`) and puts it to work in aiur. Sits on top of `logup-optimization`.

## LogUp grouping (multi-stark side, recap)

One chained-accumulator step may consume `k` consecutive messages:

```
(Π_j m_j)·(acc′ − acc) − Σ_j mult_j·Π_{j′≠j} m_{j′} = 0
```

which forces `acc′ = acc + Σ_j mult_j/m_j` exactly as `k` ungrouped steps would, but commits `⌈L/k⌉` accumulator slots instead of `L` — at the cost of raising the logUp constraint degree to about `Σ_j deg(m_j) + 1`. The group size is **circuit-local** (`CircuitInputs::lookup_group_size`, default 1) because the degree headroom depends on each circuit's message degrees; it is bounded by
`MAX_LOOKUP_GROUP = 8` and observed into the transcript shape.

## vk codec

The group size is a free per-circuit choice (it changes the constraint *structure*, not just counts), so it can't be derived: each circuit record now carries a `u8 lookup_group_size` after `max_constraint_degree`, validated against `MAX_LOOKUP_GROUP` on decode. The derived quantities become `stage_2_width = max(⌈L/k⌉, 1)·D` and `constraint_count = zeros + ⌈L/k⌉·D`. The round-trip test covers
a `k = 2` circuit.

## Lean recursive verifier

- **`SystemDeserialize.lean`** reads the group-size byte and derives `⌈L/k⌉` in-circuit by walking the lookups with a countdown (no in-circuit division). The observed shape is now **7 limbs per circuit** — `observe_shape` gained `lookup_group_size` after `stage_2_width`.
- **`Verifier.lean`**'s `logup_steps_fold` evaluates one constraint per lookup *group*, building each group's message product `P` and numerator `R` in a single pass via `R ← R·m + mult·P; P ← P·m`. At `k = 1` this reduces to exactly the previous per-lookup chained step. The wrap group still targets the next row's slot 0 plus the `is_last_row`-scaled injection.

## Branchless functions: raw arguments + `k = 2`

Lookup slots shared across branches superpose their arguments (`Σ_b sel_b·arg_b` — sound because selectors are mutually exclusive), which costs a degree: every argument becomes degree 2. A **branchless** function (single selector, no matches) has exactly one branch per slot, so the weighting is pure waste. Its arguments are now sent raw (degree ≤ 1) via `ConstraintState::gate`; the
selector-gated multiplicity alone decides whether the lookup counts.

This is witness-compatible with **no witness-side change**: the trace builder already records raw concrete arguments, active rows have `sel = 1` so the expressions agree, and padding rows have multiplicity 0, which makes the message value irrelevant — the chained accumulator rides through unchanged.

With degree-1 messages, two lookups fit in one chained step at degree 3 — within the budget the selector-gated user constraints already pay for — so synthesis sets `lookup_group_size = 2` on branchless functions with ≥ 2 lookups: **half the stage-2 accumulator columns**, plus a smaller vk (no selector-mul nodes per argument). Branching functions, memories, and gadgets keep `k = 1`.

## Testing

- `cargo test -p aiur` (incl. vk round-trip with a grouped circuit) — pass
- `lake test -- --ignored multi-stark` / `recursive-verifier` — pass; the e2e now proves through a branchless entrypoint (`fact_entry(n) = factorial(load(store(n)))`, 4 lookups, `k = 2`), so the in-circuit grouped logUp fold verifies a real grouped proof and still rejects tampering
- `lake test -- --ignored ixvm` — all 66 FFT pins green, codegen/bytecode parity green
- `lake test -- --ignored aiur` / `aiur-hashes` (blake3 + sha256, grouping-heavy) / `rbtree-map` — pass
- clippy, fmt, deny — clean


gabriel-barrett:
!benchmark aiur fresh

gabriel-barrett:

gabriel-barrett:
!benchmark aiur fresh

gabriel-barrett:

gabriel-barrett:

gabriel-barrett:
!benchmark aiur fresh

arthurpaulino (approved):

PR #527, opened 2026-07-30
multi-stark: reduce the query index per batch for the preprocessed MMCS

gabriel-barrett:
The in-circuit verifier passed the full query-index bit list and log_gmax to mmcs_verify for all four input batches. That is only correct when the batch's own max height equals the global max — always true for the stage_1/stage_2/quotient batches (they contain every circuit), but not for the preprocessed batch, whose tree only spans circuits carrying a preprocessed matrix. When the tallest circuits have no preprocessed matrix (e.g. the Array.extract_append closure: log_gmax 25 vs prep tree height 18), mmcs_fold exhausted the shorter authentication path, recomposed the leftover index bits into a nonzero cap index, and list_lookup walked off the single-digest cap — rejecting valid proofs with "no match case for value 1" (a Cons destructure hitting Nil). Small closures like Nat.add_comm dodged the bug because their prep tree height coincided with log_gmax.

Mirror Plonky3's verify_fri per-batch reduction
(index >> (log_global_max - log_batch_max), i.e. dropping the low bits of the LSB-first list) and verify against the batch's own max height (heights_max over heights_prep). The reduced-opening x-coordinates were already per-height reduced in ri_apply and need no change.

Also adds `bench-recursion-debug`, a debugging harness that proves one constant's typecheck, saves the proof/vk/claims blobs under recursion-debug/, and replays verify_multi_stark_proof from the saved blobs under the codegen'd, bytecode-interpreter, or Lean debug-interpreter backends — so a 72 s prove runs once while the verifier is iterated on.

Validated on saved proofs (native + bytecode, matching fft-costs): Array.extract_append now accepted (was rejected), Nat.add_comm unchanged; `IxTests multi-stark` and `IxTests recursive-verifier` (incl. tampered proof/claim rejections and codegen/interpreter parity) pass. aiur_multi_stark.rs regenerated via `ix codegen`; aiur_ixvm.rs unchanged.

### DIFF
--- a/Ix/MultiStark/Pcs.lean
+++ b/Ix/MultiStark/Pcs.lean
@@ -820,8 +820,15 @@ def pcs := ⟦
         let BatchOpening.Mk(rows_p, proof_p) = list_lookup(input_proof, 3);
         -- one opened base row per preprocessed matrix (BatchOpenedValuesCountMismatch)
         assert_eq!(eq_zero(list_length(rows_p) - list_length(prep_round)), 1);
-        assert_eq!(mmcs_verify(prep_commit, rows_p,
-          heights_prep(log_degrees, log_blowup, prep_indices, num_circuits, 0), idxbits, proof_p, log_gmax), 1);
+        -- The preprocessed tree's max height is over the PREP circuits only, so
+        -- it can be SHORTER than the global max when the tallest circuits carry
+        -- no preprocessed matrix. Mirror Plonky3's per-batch index reduction
+        -- (`index >> (log_global_max - log_batch_max)` — dropping the low bits
+        -- of the LSB-first bit list) and verify against the batch's own height.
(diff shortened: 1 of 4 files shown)


Ix/MultiStark/Pcs.lean
```diff
@@ -835,6 +842,18 @@ def pcs := ⟦
...
+  fn heights_max(hs: List‹G›, acc: G) -> G {
+    match load(hs) {
+      ListNode.Nil => acc,
+      ListNode.Cons(h, rest) =>
+        match u32_less_than(acc, h) {
+          0 => heights_max(rest, acc),
+          _ => heights_max(rest, h),
+        },
+    }
+  }
```
arthurpaulino, line 857: Can the `acc` parameter be dropped by using direct recursion?

arthurpaulino (dismissed):

arthurpaulino (approved):

PR #537, opened 2026-08-04
feat(aiur): Accurate whole-env sharding based on execution measurements

samuelburnham:
# Aiur sharding: measured scan-and-cut

## What this does

Shard boundaries for Aiur proving are now **measured, not predicted**. `ix shard env.ixe` executes the env's check schedule through the codegen'd Aiur kernel — the same thin-frontier `CheckEnv` claims a prover pays for — with a running cost readout, and cuts a shard wherever the measured cost reaches the RAM budget. The previous counter-based cost model (fit on ingress bytes / substitutions / def-eq counts) mispredicted real content by up to 16× in both directions and OOM'd shards it called safe; execution is the mandatory prefix of proving, so measuring it prices the prove exactly.

## Design

- **Two-term cut**: a shard ends when `RAM-per-fft · measured fft + measured record bytes` reaches the budget's usable headroom — both resources the prove actually consumes.
- **One RAM mechanism, no reactive control**: each scan worker gets a record-byte cap fixed at startup (`0.70·box RAM / workers`), enforced inside kernel execution. Fleet RAM is `workers × cap` by construction. A block that outgrows the cap alone is skipped and named in `failed.csv`, same as a kernel-rejected block — pathological content is excluded by name, not engineered around.
- **True per-shard costs**: every shard is re-priced with one cold `CheckEnv` — exactly the claim its prove executes — so manifest costs carry no scan-side inflation. The `.ixes` manifest (format v2) stores them per shard; `ix prove` orders heaviest-first from it.
- **Lazy witness fault-in**: the witness `IOBuffer` faults env bytes on demand, so witness RAM tracks the touched set, not the shipped closure — for scan and prove alike.
- Parallel scan via chunked schedule + work stealing; `--fail-fast` (default) / `--no-fail-fast` uniform across `ix shard` / `check` / `prove`; whole-env rows in the `aiur-shard` bench backend (scan time, peak RSS, shards, fft totals).

## Validation (250 GiB budget, 32-core / 250 GB box)

| env | blocks | scan wall | peak RSS | shards | total fft | max pred. prove RSS |
|---|---|---|---|---|---|---|
| Init | 51,003 | 152 s | 91 GiB | 40 | 1,719 BFFT | 165.4 GiB |
| InitStd | 87,847 | 255 s | 127 GiB | 76 | 3,330 BFFT | 159.7 GiB |
| Lean | 152,596 | 383 s | 123 GiB | 108 | 5,086 BFFT | 164.6 GiB |

All blocks covered (zero skips), partitions deterministic, every predicted prove RSS under budget. Scan cost is ~2% of the pack's prove cost. Measured prove check: shard predicted 76.1 GiB / 60.5 s proved at 74.8 GiB / 59.4 s.

## Usage

```bash
ix shard Init.ixe --max-ram 250 --out Init.ixes   # scan → manifest + costs sidecar
ix check --ixe Init.ixe --ixes Init.ixes          # check shards
ix prove --ixe Init.ixe --ixes Init.ixes          # prove, heaviest-first, resumable
```

The shard-pipeline E2E runs in `lake test -- --ignored ixvm` (also in PR CI).

## Known limits

- FLT/Mathlib contain a handful of pathological blocks (kernel-rejected or with prove costs beyond any single-box budget); scans name and skip them rather than rescue them, and the coverage gate reports the exclusions.
- The fft→RSS line is fit on 13 measured Init proves with 3 spot checks on other envs (−1.7% / +2.9% / +6%); a broader stratified prove sample is follow-up work.


samuelburnham:
Superseded by #601

PR #541, opened 2026-08-07
Universe-level canonicalization and Decompilation bugfix

johnchandlerburnham:
# Universe-level canonicalization (canonicity §10.6), alias-provenance metadata (§10.5), and whole-Mathlib scaling of the pure-Lean validator

This branch lands three related bodies of work, culminating in the §10.6 universe-level quotient: content addresses now coincide with the kernels' semantic level equality, with source spellings preserved losslessly in metadata. Whole-Mathlib validation is green in both implementations, strict everywhere, with the two compilers byte-identical.

## Part I — Canonicity §10.5: metadata name provenance (prerequisite fixes)

The whole-Mathlib byte-parity investigation surfaced a 47-byte divergence (`Quiver.FreeGroupoid.redStep.{rec,casesOn,recOn}`): the kernel's WHNF intern-collapsed alpha-identical wrapper defs (`Paths`/`Symmetrify`) to first-interned spellings, making synthesized metadata name choices schedule-dependent.

- **§10.5 provenance rule** (spec + implementation): synthesized occurrences inherit the spelling of the source occurrence they derive from — never a class-representative choice made at emission. Kernel-cache state no longer outlives the block (`compile: block-scope kernel contexts`).
- **Root cause worth remembering**: Rust `KExpr::hash_key()` is an intern UID while Lean `Tc.KExpr.addr` is a content digest — the source-name hint map keyed by `hash_key` had never matched (dead since birth). Fixed with a name-erased structural content key mirroring the two kernels' induced equivalence rather than their accessor spellings.
- **Level-aware nested-aux identity** (`aux_gen`): three sites still matched auxes by (family, term-specs) only, collapsing distinct universe instantiations (`DedupM`/`UnivM` fixtures); all three now key on levels with an exact-then-insensitive two-pass, mirrored Rust↔Lean.

## Part II — Whole-Mathlib scaling of `ix validate-lean`

The pure-Lean validator previously could not complete Mathlib (several independent >100 GiB blowups). Now it completes in ~90 min at <100 GiB:

- **Streaming compile oracle**: proof bodies stream through canon and are never materialized (hybrid: code kinds stay resident); phase-5 oracle is per-name digests instead of a whole-env canon copy.
- **Byte-backed constant storage**: compiled constants held as serialized bytes, not object graphs.
- **Streaming serde gate**: per-unit parse→reserialize→compare with gapless span coverage (`deEnvVerifiedLazy`) instead of whole-env materialization.
- **Streaming meta roundtrip**: per-chunk materialize→ingress→egress→drop; the merged whole-env MetaEnv never exists.
- **The big one**: `Tc.canonExpr` and derived `BEq` exponentially unfolded pointer-shared DAGs on egressed constants (multi-GiB transients from 2 KB constants; 5.6 h of comparison). Both are now pointer-memoized (`ptrAddrUnsafe`, soundness argument in-module): 73.4 GiB/460 s → 5.1 GiB/13.2 s on the bisection slice, 41 s for all of Mathlib.
- **Streamed phase output**: `validate-lean` now prints each phase's section heading + result the moment it completes, flushed (block-buffered end-only output twice destroyed the evidence of killed runs), matching `ix validate`'s format.

## Part III — Canonicity §10.6: the universe-level quotient

**The quotient.** Two levels are identified exactly when the kernels' semantic equality (`univEq`) holds — the endpoint quotient: content addresses coincide with kernel identity. Spellings are presentation. Declaration-level parameter *list* order stays structural; only spelling *inside* level expressions is quotiented (`max u v` = `max v u`; `(max u v)+1` = `max (u+1) (v+1)`; the WF-recursion `eq_def` shapes).

**Canonical representative.** `canonUniv = linearize ∘ subsumption ∘ normalizeAux` — the kernels' Géran comparison form, linearized back into a term by **per-atom gate inversion** (each atom self-strips gates its value dominates; gate order recovered greedily outermost-first; formerly open detail O1, settled empirically). Properties P1–P6 (idempotence, roundtrip-fixpoint, mk*-fixpoint, kernel-oracle soundness, Rust↔Lean byte parity, mk* absorption) are pinned exhaustively over all ≤7-node terms plus 50k quickcheck, in both languages, with FFI cross-checks.

**`univEq` is now exact** (option (b)): the normal-form comparison ignores empty subsumption entries (as `normLevelLe` always did) in all three kernels (Rust, Lean Tc, IxVM). Before this, 3 of 3,253,373 whole-Mathlib entries were distinguished from their semantic equals.

**Restoration metadata.** Per-occurrence `ConstantMeta.univPatches` (arena-node-keyed; full argument lists for const occurrences) + a `metaUnivs` extension table under the virtual-index contract, as a fourth wrapper vector in both serializers (+ FFI codec, diff labels, generators, fixtures). Table-keyed patching is unsound (canonicalization dedups distinct spellings onto one entry — 79,088 Mathlib constants contain a collision); arena keying is exact because expression identity is spelling-injective. One structural subtlety mirrored everywhere: a surgered call-site head's arena root is unreachable during replay, so its patch is cloned onto the callSite node root.

**Kernel contract.** Anonymous ingress never reads patches (they influence no hash and no judgment). Meta ingress decorates occurrence nodes with original spellings — folded into `metaAddr` only, never `addr` (anon/meta parity preserved; checking never sees spellings) — sourced from patches with the stage-1 mk*-rebuild rule as the patchless fallback. Meta egress replays decorations. Comparators were never weakened; with canonical tables the anon roundtrip dropped its `reduceIxonUniv` modulo and is now **strict**.

**Execution order** (per plan): stage 1 (decorations, no format change) → census probe → stage 2 Rust-first (compiler + kernel end-to-end on Rust-only gates) → Lean mirror against the cross-compiler gates → format break with a "pre-normal-levels .ixe; recompile it" parse hint → primitive-pin regeneration everywhere (`prim_addrs.rs`, `Ix/Tc/Primitive.lean`, IxVM address literals — 56 pins; LEON pins unchanged) → Aiur codegen regeneration + FFT cost re-pins (66 pins, all within ±0.3%, every functional/parity check green).

**Probe** (`dump_reducible_univs`, kept as a permanent census tool): whole-Mathlib blast radius was 373,799 Géran-noncanonical entries in 134,929 constants (~1.04 M occurrences, ~10.9 MB patches, 0.34%), 84% dependent closure. Post-regen artifacts: **`Géran-noncanonical: 0`, collision constants 0, src == canonical bytes.**

### Bugs found and fixed along the way

- **Egress table-pairing hazard (measured, then designed away)**: the kernel-ixon roundtrip pairs rebuilt constants with original metas, but rebuilt first-use tables diverge from preseed-sorted originals on 61–98% of bodies — previously benign only because *no* metadata referenced table index space. `univPatches` would have been the first. Egress now preseeds each rebuilt univ table verbatim from the original constant (pairing exact by construction, debug-asserted).
- **Pointer-keyed memo vs ephemeral metas** (caught by a flaky `Std.DHashMap.Raw.WF` re-run): demoted metas re-parse per access, so ctor-window extension univs were sole-owner allocations; freed addresses collided in the decompiler's `*const Univ`-keyed level memo, substituting arbitrary stale spellings allocator-dependently. Fixed by invalidating the memo at the window; regression-pinned with a multi-ctor patched-inductive fixture. (The Lean decompiler's per-constant `withFreshBlock` design is immune by construction.)
- **Ctor extension offset**: per-ctor `metaUnivs` must install at the *primary* table length, not the parent-extended length (latent until extensions became non-empty).
- V3 preseed-finality tripwires in both compilers (primary table growth after preseeding would silently shift virtual patch indices).

## Validation

| Gate | Result |
|---|---|
| `ix validate` (Rust 8-phase), whole-Mathlib | **0 failures** (736,624) |
| `ix validate-lean` (pure Lean 5-phase), whole-Mathlib | **0 failures**; phase 3 strict (647,052); phase 4 = **714,346 spellings / 0** (closes the 111 standing `levels differ` findings); phase 5 all digest-identical |
| Rust kernel typecheck, whole-Mathlib | 736,624/736,624 |
| `compile-lean --rust-check`, RedStep + Mathlib | **ALIGNED** — 3,155,562,665 bytes byte-identical |
| kernel-ixon-roundtrip / rust-compile / validate-aux | 0 / 150,396 · 0 / 228,770 (incl. 577 MB serde) · 0 |
| tc-unit / tc-roundtrip / tc-ingress-meta / decompile-diff / aux-gen-diff / prim-addrs / ixvm | all green |
| cargo workspace | 1,249 tests, clippy clean |

Docs: §10.6 rewritten as live spec (linearizer + exact `univEq` + patch contract), §12.4 worked example, §17.9 landed record; `Ixon.md` univ-table invariant + ConstantMeta wrapper layout. `BENCHMARKS.md` refreshed (regenerated artifact sizes, Mathlib timings, previously-TBD validate-lean column).

**Follow-ups** (tracked in §17.9): kernel-side univ-table canonicity enforcement at ingress (reject, never silently canonicalize; all three kernels + foreign-`.ixe` policy); Tc Verify-layer proofs of P1/P2/P4.

Format break: pre-existing `.ixe` artifacts are invalidated (parse error with a recompile hint); regenerate-everything was the adopted policy (D4).


samuelburnham:
!benchmark compile decompile

arthurpaulino (approved):

PR #544, opened 2026-08-07
multi-stark quotient lde bump

gabriel-barrett:
# Quotient commit from coefficient slices (multi-stark bump) + FFT cost model update

Bumps multi-stark to [`d85ba29`](https://github.com/argumentcomputer/multi-stark/commits/quotient-commit-from-coefficients), which changes how the prover commits the quotient, and updates the Aiur FFT cost statistic (#539) to model the new transforms.

## multi-stark side

The quotient is the only committed data born in coefficient form, yet the prover treated it like a witness trace: DFT the coefficient slices onto the trace domain, then let `Pcs::commit`'s `coset_lde_batch` immediately inverse-DFT them right back. The DFT/iDFT pair cancels exactly, so it was `2·q·D` wasted size-`n` transforms per circuit.

Two commits:

1. **`prover: commit the quotient straight from its coefficient slices`** — build the committed LDE directly from the slice coefficients as **one zero-padded size-`B·n` DFT** per circuit (fold the `GENERATOR` shift into the coefficients, pad, `dft_batch`; the transform's native bit-reversed storage is exactly the layout `TwoAdicFriPcs::commit` stores), and commit through `Pcs::commit_ldes`.
An earlier per-coset variant (B separate `coset_dft_batch` calls) did fewer butterflies but *regressed* the quotient span 5.59s → 6.88s: the default `coset_dft_batch` pays a matrix clone, a fully **serial** shift-scaling pass, and two bit-reversal passes per call, which dominates at real heights (n ≈ 2²⁵) while staying invisible in small-height microbenchmarks.
2. **`prover: fuse the quotient iDFT tail into the slicing gather`** — the iDFT post-passes (bit-reversal materialization, serial row swap, serial coset-unscale) and the serial slicing loop collapse into one parallel gather: the coset-unscale `g^{−(k·n+r)}` cancels the LDE pre-scale `g^{r}` exactly, leaving a single constant `N⁻¹·g^{−k·n}` per slice.

The committed matrix is value-identical to the old path — transcript, proof bytes, and verifier all unchanged. Two unit tests pin the equivalences (against `coset_lde_batch ∘ coset_dft_batch`, and against the naive iDFT + slice + pre-scale composition).

### Benchmarks (`bench-typecheck --ixe init.ixe --consts String.split --texray`, 64 cores)

| multi-stark rev | stark/quotient | prove total | proof bytes |
|---|---|---|---|
| `5023fee` (main) | 5.59s | 32.1s | 23139426 |
| **`d85ba29` (this PR)** | **4.97s (−11%)** | **31.0s** | 23139426 |

## ix side

The per-circuit FFT statistic (#539) modeled the old quotient path. Updated to the new transforms:

```text
(B+1)·(m + s2 + q·D)·F(h) + D·F(q·h) + q·D·F(h)
→  (B+1)·(m + s2)·F(h)    + D·F(q·h) + q·D·F(B·h)

The quotient chunk columns lose the (B+1) commit transforms (commit_ldes runs no FFT) and the size-h rebase DFT, and pay the padded size-B·h DFT instead. Trace commits still go through coset_lde_batch and keep their (B+1)·F(h) shape.

- All 66 kernel-check pins and the shard pipeline pin regenerated from one lake test -- --ignored ixvm run (all execution/parity tests pass; real-circuit costs drop ~2% uniformly).
- Drive-by: the shard pin's message text had drifted from its asserted value (10817625733 vs 10_785_479_733); both now carry the same number.
- The aiur-cost hand-computed value moves 1424 → 1488: per chunk column the padded transform costs F(B·h) against the old (B+2)·F(h), which only wins for log₂(h) > 4 — a deliberate small-height loss at the test's h = 8; every real circuit clears it.


gabriel-barrett:
!benchmark aiur fresh

gabriel-barrett:

arthurpaulino (approved):

PR #546, opened 2026-08-11
IxVM def-eq: rebase open pairs to their minimum loose index

arthurpaulino:
`Algebra.kerTensorProductMapIdToAlgHomEquiv._proof_1` (mathlib) never finished: not in 300s natively, against 593ms for the reference kernel on the same subject. Instrumenting both kernels showed the pair space was NOT the problem — IxVM visits 15,085 distinct (a,b) pairs against the reference's 5,941 — but each pair was being recomputed once per context spelling: 2.48M distinct (a,b,types) def-eq keys, 164x per pair, and the whole delta cascade re-ran under each key (26.1M delta unfolds against the reference's 11,544, concentrated on the DFunLike.coe / *.toFun instance-projection towers).

The prefix trim cannot canonicalize these pairs. The proof binds its section variables (types, rings, instances) OUTERMOST, so a tower pair under the inner binders has a loose range spanning the whole context; the trim keeps every inner frame, and every comparison site spells those (unreferenced) inner frames differently. The reference is immune because its def-eq opens binders with fvars: compared bodies have lbr = 0 and its context key collapses to empty, so tower obligations cache on content alone.

Fix, the de Bruijn equivalent of that key insensitivity: `g = min` of the two sides' minimum loose indices — no loose BVar below `g` exists in either side, so lowering both by `g` and dropping the `g` innermost frames is a bijection on derivations (context strengthening by unreferenced frames, the same argument as the existing base=0 -> Nil arm). Def-eq returns a bit, so nothing lifts back, and one edit site covers every downstream circuit. Depth-shifted copies of the same tower also lower to the same hash-consed pair, merging pairs across binder depths. New `expr_glb` (memoized min-loose-index with an lbr fast path) supports it; the lowering is the pre-existing `expr_lower`.

A depth threshold (rebase only past base > 8) was measured and rejected: it recovered only a third of the overhead on binder-heavy Init proofs for no gain on the mathlib constant.

Measured: the constant's subject-only check completes in 13s (4.7e10 FFT) where it previously never terminated, and its FULL 8,753-constant closure now checks in 77s. All 34 previously-fixed mathlib constants pass full-closure in 40s total. Guard set flat: goCache_Inv_of_Inv 112s, goCache.eq_def 42s, ofBitVec_not 3s, ofDays 6s. Both suites green. Pins re-harvested: +0.0% to +6.4% (the worst are binder-heavy proofs paying the glb walk with nothing to rebase — the price of the unconditional form). New pinned fixture `deepRebase` minimizes the outer-reference telescope shape.


arthurpaulino:
!benchmark aiur fresh

gabriel-barrett (approved):

PR #557, opened 2026-08-13
perf/inline blake3 round

gabriel-barrett:
Inline `blake3_compress_inner_j`. This has the effect of increasing the total width, though not at the same proportion as what we gain on efficiency, for the lookup itself is costly, especially when the input arguments are big.

Eventually we will have to decrease the proof size of the recursive verifier, but there are ways around it see #551


gabriel-barrett:

gabriel-barrett:

gabriel-barrett:
!benchmark aiur fresh

gabriel-barrett:

gabriel-barrett:
!benchmark aiur fresh

gabriel-barrett:

gabriel-barrett:
!benchmark aiur fresh

PR #560, opened 2026-08-14
IxVM: fix false rejections and pathological kernel reductions

arthurpaulino:
Full-environment IxVM sweeps exposed two classes of kernel bugs: valid declarations rejected by canonical/positivity checks, and open def-equality or inference workloads that failed to converge in practical time.

Correct the recursor canonical checker to reconstruct each mutual member from its own index telescope and only permit Prop large elimination for solo blocks. Skip strict-positivity checking for unsafe inductives, matching Lean and the Rust kernel.

Canonicalize open inference memo keys by rebasing subjects to their minimum loose index, then lift inferred types back into the caller context. This prevents irrelevant inner binder frames from poisoning cache reuse.

Keep lazy def-equality reduction bounded and no-delta between steps. Reuse the current lazy budget instead of recursively restarting application congruence, recognize compact Nat offsets, preserve K/projection reductions under no-delta WHNF, and retain a last-resort full-WHNF path for representations that cannot otherwise converge.

Add focused fixtures for the mutual-recursors, unsafe-inductive, inference-rebase, large-offset Char, and indexed Time-cast failure shapes. Regenerate the native IxVM and repin all kernel and shard FFT costs after rebasing onto main.

Validation: lake exe ix codegen; lake build IxTests; lake test -- --ignored ixvm (187.06s, all execution, FFT, exploit, and bytecode/native parity checks passed).


arthurpaulino:
!benchmark aiur fresh

samuelburnham (approved):

PR #561, opened 2026-08-15
recursive verifier minimization

gabriel-barrett:
## Summary

Reduce the width and FFT cost of the recursive verifier by restructuring BLAKE3 compression, removing unnecessary circuit boundaries, narrowing several verifier operations, and representing large values more efficiently across circuit boundaries.

For the measured `factorial(5)` recursive-verifier workload, total circuit width decreases from **11,239 to 7,784** (**30.7% reduction**), while the circuit count decreases from **202 to 146**.

### Stage BLAKE3 compression across rows

Previously, all seven BLAKE3 rounds were unrolled into one wide, straight-line circuit. This PR changes `blake3_compress` into a staged recursive circuit:

- `blake3_compress_init` assembles the initial state and is always inlined.
- Each recursive step executes one compression round.
- The message permutation is applied between rounds.
- Stage 7 performs the feed-forward XOR and returns the digest.
- The BLAKE3 wrapper is inlined at its call sites.
- The chunk-tree layer is threaded by pointer instead of as a 33-column value.

This changes the compression circuit from:

- `blake3_compress`: 1,080 columns
- `blake3_compress_inner_j`: 1,192 columns

to:

- staged `blake3_compress`: **925 columns**

The tradeoff is additional height: **654 → 5,544 rows**, in exchange for substantially lower width.

Measured end-to-end:

- Total width: **11,239 → 9,711** (**13.6% reduction**)
- FFT cost: **5.51e8 → 5.05e8** (**8.4% reduction**)

### Inline single-use and pure-wiring helpers

Inline approximately 40 single-call-site functions across:

- Proof, system, and claims deserialization
- Fiat–Shamir
- OOD verification
- PCS/FRI verification
- Merkle verification
- Digest conversion and transcript helpers

A dedicated circuit has fixed input, output, selector, quotient, and lookup-communication costs. For single-use branchless helpers, those costs provide no benefit.

This reduces:

- Total width: **9,711 → 8,697**
- Circuit count: **202 → 158**

Additional single-site helpers are subsequently inlined or folded into their callers, bringing the final circuit count to **146**.

### Narrow challenger sampling

- `ch_sample8` now pops eight bytes directly and flushes at most once.
- `ch_sample_bits` decomposes only the low four sampled bytes.
- The full eight-byte draw is preserved to maintain Fiat–Shamir alignment.
- Proof-provided logarithmic values are explicitly bounded to 32 bits.

This removes unnecessary per-byte sampling and 64-bit decomposition work.

### Compute the two-adic generator iteratively

Replace the 33-arm `two_adic_gen` table with a memoized squaring chain starting from the maximal \(2^{32}\) generator:

```text
g_k = g_{k+1}²
```
This reduces the corresponding circuit width from 73 to 17 columns.

### Narrow verification-key deserialization

- Compute gslots without branching.
- Inline read_sys_circuit into read_sys_circuits_n.
- Compute repeated query heights once and reuse the result.

Together with the challenger and generator changes:

- Total width: 8,697 → 8,520
- Circuit count: 158 → 155

### Thread Merkle digests by pointer

A digest occupies 32 columns when passed by value, while a pointer occupies one.

This PR introduces DigestP = &Digest and uses pointers throughout:

- Merkle caps
- Authentication paths
- Compression inputs and outputs
- The running node in mmcs_fold

Digest bytes are loaded only where their contents are actually consumed.

Because Aiur memory is content-addressed, digest pointer identity is equivalent to digest value identity. Existing cross-query compression memoization is preserved.

Notably:

- mmcs_fold: 297 → 49 columns
- Total width: 8,520 → 8,166
- FFT cost at 50 queries: 3.583e9 → 3.426e9 (4.4% reduction)

### Compare Merkle roots by pointer

Since digest memory is content-addressed, the recomputed root can be checked against the cap entry using pointer equality instead of loading and comparing two 32-byte digests.

This reduces mmcs_verify from 97 to 22 columns.

### Pass logUp values directly

ood_composition previously read eight logUp coordinate values back from the public-values list immediately after constructing it.

These values are now passed directly as arguments, while the list remains available for public-node lookups.

This reduces ood_loop from 285 to 251 columns.

### Simplify proof and node deserialization

- Destructure the single-constructor Proof directly.
- Inline prep_onto.
- Fold read_node into read_nodes_n.
- Duplicate the recursive continuation across mutually exclusive node-tag arms, allowing those arms to share columns.

This removes additional single-use circuit boundaries while preserving the existing wire format.

### Pack public digest inputs

The verifier previously exposed the verification-key and claims digests as 64 individual byte-valued public inputs.

This PR packs each group of four little-endian bytes into one field element. The packing is injective over Goldilocks because (2^{32} < p).

As a result:

- Digest public inputs: 64 → 16 columns
- Entrypoint width: 782 → 734
- Total width: 7,951 → 7,903

> [!IMPORTANT]
> This changes the public-input wire format of verify_multi_stark_proof.
> External callers should construct the new representation through
> verifierPubInput.

### Observe claims bytes directly

The serialized claims stream is already byte-identical to the corresponding Fiat–Shamir transcript segment.

Instead of parsing the claims and then serializing them again for observation, the verifier now observes the original claims bytes verbatim.

This also:

- Removes claims_onto, claims_each_onto, and claim_vals_onto.
- Removes two now-unused list_length_u64 monomorphs.
- Observes sampled lookup and fingerprint challenges using one 16-byte concatenation each.

Final measured totals:

- Total width: 7,903 → 7,784
- Circuit count: 149 → 146

## Benchmarking support

Add --interp to bench-typecheck, allowing execution through the generic Aiur bytecode interpreter.

This supports fast iteration on Ix/IxVM/*.lean without requiring IxVM code generation and a Rust rebuild after every change.

The option applies to:

- Phase-one execution
- Proving witness generation
- Recursive execution
- Recursive proving

FFT-cost measurements remain executor-independent. Execution-time measurements should not be compared directly between interpreter and generated-code modes.

## Validation

The following were run across the commits in this PR:

- Multi-STARK self-tests
- Recursive-verifier end-to-end acceptance tests
- Generated-code/interpreter parity tests
- Tampered-proof rejection tests
- Aiur Cargo tests
- Aiur proving and cross-execution tests
- Formatting and Clippy checks

Generated IxVM and multi-STARK Rust code has been updated.


gabriel-barrett:

gabriel-barrett:

gabriel-barrett:
!benchmark aiur fresh

gabriel-barrett:

gabriel-barrett:
!benchmark recursive fresh

arthurpaulino (approved):

PR #571, opened 2026-08-19
Prop-mutual `.below` families — canonical rec order + casesOn regeneration

johnchandlerburnham:
Fixes the known-open finding from the mega-PR's real-library benchmark:
`ix check-rs` rejected 25 constants of HaskellSpec's env — every
`.below` auxiliary of its Prop-valued *mutual* inductive predicates
(dict/pat/type/bind/exp/gde… families; `main` rejected a strict
superset of 37 before the aux-alias fix). After this branch the
HaskellSpec env and the whole CatalogReal catalog kernel-check clean.

## Two root causes, both compile-side

**1. The below-rec block ignored §6.2.** Prop-mutual predicates get an
IndPredBelow-generated *mutual-inductive* `.below` family, and its
recursor block was compiled through the keyless `compile_aux_block` —
so `sort_consts` picked its own structural permutation. Sibling
below-*recursors* first differ at index/motive shape while sibling
below-*inductives* first differ at ctor content, so the two blocks'
canonical orders can diverge — and the kernel's
`populate_recursor_rules_from_block` pairs `rec_block[i]` with
`flat[i]` positionally (canonicity §6.2, verified loudly). The main
recursor block already passes a class-order key for exactly this
reason; the below-rec block now does too.

**2. Lean's authored `X.below.casesOn` dies once the rec is
canonical.** Its value applies `X.below.rec` with motives in *Lean's*
member order; ix regenerates that rec with the canonical motive
layout, so the authored wrapper becomes genuinely ill-typed in the
compiled env (the AppTypeMismatch and both universe-param-count
rejections). Ordinary families never hit this because ix regenerates
their `casesOn` from the canonical rec (Phase 1b/2b); the below path
now does the same.

**Decompile had to learn the new shape:** `classify_aux_gen` rooted
`X.below.casesOn` at `X.below` — an aux inductive, not a family — so
the member joined no processed block and silently vanished from
decompiled envs; it now roots at the family like `X.below.rec`. And
since the wrapper's rec only exists after the below-rec phase, its
regeneration is deferred to a new Phase 3b fed by the *generated*
`RecursorVal`s (the Lean driver's `wrapPhase` gains a rec-resolver
parameter; the Rust side factors Phase 1b's body into
`regen_one_cases_on`).

## Commits

1. **kernel: emit gated canonicity/recursor dumps on stderr** — the
   `IX_RECURSOR_DUMP` / alignment-failure / rule-diff dumps logged via
   the `log` facade with no backend installed in the CLI; every
   breakdown the kernel's own error hints point at was silently
   dropped. Now stderr, like the compile-side mirror dumps.
2. **compile: align the below-rec block with its inductive block;
   regenerate below.casesOn** (mutual.rs + AuxGen/CompileAux.lean).
3. **decompile: regenerate below.casesOn against the canonical
   below-rec (Phase 3b)** (decompile.rs + DecompileDriver.lean +
   CallSiteSurgery.lean classifier).
4. **tests: `BelowPredicate` corpus fixture** — EvenP/OddP mutual Prop
   predicates + mutual theorems by structural pattern-matching (the
   IndPredBelow trigger), permanently swept by validate-aux /
   aux-gen-diff / decompile-diff and the C5 qualified corpus.

## Verification

| gate | result |
|---|---|
| HaskellSpec standalone (`CompileHS.lean`) | 25 rejections → **64,051/64,051 kernel-clean** |
| CatalogReal (`ix catalog --audit` + `check-rs`) | **208,583/208,583**, audit still green (19,584 address-preserving), root `957e5f8cfc35…` |
| full `lake test` | 2,670 pass (fixture included) |
| `validate-aux` | 0 failures |
| `aux-gen-diff` | Rust and Lean drivers byte-identical (6,247,773 B), serial + parallel |
| `decompile-diff` | all gates PASS (300 plans; aux-family recovers the 19 dropped below-casesOn) |
| mathlib Rust compile | 736,624 consts → **3,155,811,864 B**, `ix validate` **0 failures** |
| mathlib ALIGNED (`compile-lean @16 --rust-check`) | **byte-identical with Rust**, and `cmp`-identical with the on-disk artifact |
| mathlib `validate-lean` (5-phase, @8) | phases 1–4 PASS (phase 1 reproduces the canonical bytes; phase 4 zero findings, `skippedAux` +99 = the below-casesOn constants now on the aux track); phase 5 decompile in flight at PR time |

## ⚠️ Canonical bytes move (+249,024 at mathlib scale)

`X.below.casesOn` storage changes from standalone authored defs to
regenerated aux-block members across every Prop-below family
(`Nat.le`, `Relation.TransGen`, `Acc`, `List.Perm`, …). Same wire
format, different compiler output — all artifacts regenerate.
**Open decision:** bump `Env::VERSION` to 2 per the policy sentence
("any change to serialized bytes bumps VERSION"), or fold into the
still-unreleased v1 from the base PR. One-integer change either way
(serialize.rs + Ix/Ixon.lean + docs/Ixon.md).

## Downstream

TruthMinesLib should re-pin past this branch's merge (not just the
base PR): HaskellSpec is one of their 81 libraries, and their
kernel-check gates over any catalog containing it hit exactly the 25
pre-fix rejections.

### DIFF
--- a/Ix/CallSiteSurgery.lean
+++ b/Ix/CallSiteSurgery.lean
@@ -79,7 +79,17 @@ def classifyAuxGen (name : Name) : Option (AuxKind × Name) :=
     else if s1 == "recOn" || s1.startsWith "recOn_" then
       some (.recOnAux, p1)
     else if s1 == "casesOn" || s1.startsWith "casesOn_" then
-      some (.casesOnAux, p1)
+      -- X.casesOn / X.casesOn_N or X.below.casesOn. The below wrapper
+      -- roots under the FAMILY (like `X.below.rec` above) so it joins
+      -- the family block's aux members and the decompiler's Phase 3b
+      -- can regenerate it against the canonical below-rec.
+      match p1 with
+      | .str gp ps _ =>
+        if ps == "below" || ps.startsWith "below_" then
+          some (.casesOnAux, gp)
+        else
+          some (.casesOnAux, p1)
+      | _ => some (.casesOnAux, p1)
     else if s1 == "below" || s1.startsWith "below_" then
       some (.below, p1)
     else if s1 == "brecOn" || s1.startsWith "brecOn_" then
--- a/Tests/Ix/Compile/Mutual.lean
+++ b/Tests/Ix/Compile/Mutual.lean
@@ -624,4 +624,33 @@ end ThreeMember
 
 end AuxOwnership
 
+-- Prop-valued mutual inductive predicates whose `.below` auxiliaries form
+-- a generated mutual-inductive family (IndPredBelow) — the HaskellSpec
+-- `dict`/`pat`/`type` shape. The mutual theorems by structural
+-- pattern-matching force Lean to generate `EvenP.below`/`OddP.below`
+-- (plus `.below.casesOn`, defined via `.below.rec`), exercising:
+-- (a) the below-rec block's class-order key — its storage order must
+--     follow the below inductive block (canonicity §6.2), and
+-- (b) the `.below.casesOn` regeneration against the canonical below-rec
+--     (compile + decompile Phase 3b; Lean's authored wrapper applies
+--     motives in Lean's member order and would be ill-typed against the
+--     canonical rec).
+namespace BelowPredicate
+mutual
+  public inductive EvenP : Nat → Prop where
+    | zero : EvenP 0
+    | succ : {n : Nat} → OddP n → EvenP (n + 1)
+  public inductive OddP : Nat → Prop where
+    | succ : {n : Nat} → EvenP n → OddP (n + 1)
+end
+
+mutual
+  public theorem evenp_nonneg : {n : Nat} → EvenP n → 0 ≤ n
+    | _, .zero => Nat.le_refl 0
+    | _, .succ h => Nat.le_trans (oddp_nonneg h) (Nat.le_succ _)
+  public theorem oddp_nonneg : {n : Nat} → OddP n → 0 ≤ n
+    | _, .succ h => Nat.le_trans (evenp_nonneg h) (Nat.le_succ _)
+end
+end BelowPredicate
+
 end Tests.Ix.Compile.Mutual
--- a/Ix/AuxGen/CompileAux.lean
+++ b/Ix/AuxGen/CompileAux.lean
@@ -557,7 +557,53 @@ def compileBelowRecursors (belowIndcs : Array MutConst) (maps : AddrMaps)
     belowRecs := belowRecs.push (.recr rv)
 
   if !belowRecs.isEmpty then
-    compileAuxBlock belowRecs maps
+    -- The below-rec block's storage order must align with the below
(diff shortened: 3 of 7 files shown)


Ix/CallSiteSurgery.lean
```diff
@@ -79,7 +79,17 @@ def classifyAuxGen (name : Name) : Option (AuxKind × Name) :=
...
       some (.recOnAux, p1)
     else if s1 == "casesOn" || s1.startsWith "casesOn_" then
-      some (.casesOnAux, p1)
+      -- X.casesOn / X.casesOn_N or X.below.casesOn. The below wrapper
+      -- roots under the FAMILY (like `X.below.rec` above) so it joins
+      -- the family block's aux members and the decompiler's Phase 3b
+      -- can regenerate it against the canonical below-rec.
+      match p1 with
+      | .str gp ps _ =>
+        if ps == "below" || ps.startsWith "below_" then
```
arthurpaulino, line 88: Casing on `Name` fragments is so uncomfortable... but I don't see any better method at the moment

Tests/Ix/Compile/Mutual.lean
```diff
@@ -624,4 +624,33 @@ end ThreeMember
...
+    | succ : {n : Nat} → EvenP n → OddP (n + 1)
+end
+
+mutual
+  public theorem evenp_nonneg : {n : Nat} → EvenP n → 0 ≤ n
+    | _, .zero => Nat.le_refl 0
+    | _, .succ h => Nat.le_trans (oddp_nonneg h) (Nat.le_succ _)
+  public theorem oddp_nonneg : {n : Nat} → OddP n → 0 ≤ n
+    | _, .succ h => Nat.le_trans (evenp_nonneg h) (Nat.le_succ _)
+end
```
arthurpaulino, line 653: Confirmation question: adding these here is enough to exercise the fix, right?

arthurpaulino (approved):

Tests/Ix/Compile/Mutual.lean
johnchandlerburnham: Only partly, it exercises the `casesOn` path, but not the `.rec` order-divergence. I will try to add a better fixture for that

Ix/CallSiteSurgery.lean
johnchandlerburnham: yeah agreed, I think it's okay for now because this is elaboration-level, but I'm going to try to figure out a better alternative

Tests/Ix/Compile/Mutual.lean
johnchandlerburnham: If you want something to test ixvm on in the interim you can look at some of the stuff I've added to https://github.com/argumentcomputer/ix/tree/main/Benchmarks, e.g. you can directly test on https://github.com/argumentcomputer/ix/blob/main/Benchmarks/CatalogReal/CompileHS.lean

arthurpaulino: yeap, going through those now

PR #572, opened 2026-08-19
Lean v4.33.0 consolidation + module-mode full-content compile fix

johnchandlerburnham:
One PR consolidating the v4.33 toolchain update with the fix for the
module-mode environment regression that is currently failing main's
`Ix.Tc verification and parity` job (the `tc-pins` / `tc-accel-diff`
reds introduced by #569).

## Commit 1 — merge: `update/lean-v4.33.0` × post-#571 main

Supersedes **both** previous v4.33 efforts: `update/lean-v4.33.0`
(fork #568, 20 commits — taken wholesale) and `jcb/lean-v4.33.0-rebase`
(fork #562 — fully obsolete: its `patches/lean4lean-v4.33-lint-clean.patch`
landed upstream in the fork pin `4844eda4` this branch already uses, and
its ports were re-authored on the newer branch).

Only two textual conflicts, resolved as:

- `Tests/Main.lean` — union: #569's four new suites in, the branch's
  deliberate `ByteArray`-suite removal stands.
- `Ix/Meta.lean` — #569's module-header honoring resolved **into the
  corrected semantics** (commit 2's subject); the branch's
  `enableInitializersExecution` line (a v4.33 requirement) kept.

Merge fallout fixed in the same commit: the catalog smoke test's
`CatalogSpec` gains a `Plausible` entry ahead of `LSpec` — the v4.33
workspace pins LSpec with a new `plausible` dependency, and the
cataloguer is fail-closed about uncatalogued packages (and replays
dependencies before dependents).

Inherited from the branch (pre-existing, unchanged — see `HANDOFF.md`):

- ⚠️ **Blake3 pin is an unmerged branch** (`native-decide-dynlib`,
  `730f910a`). Do not merge this PR until that branch merges to Blake3
  main; then re-pin.
- `lake test -- --ignored ixvm` OOMs in the Aiur toplevel build
  (42–79 GB; diagnosis in HANDOFF §2.3). Not addressed here.
- `IxTcVerify` remains excluded from the `--wfail` lint driver
  (pre-existing decision; 331 ix-own warnings).

## Commit 2 — fix(meta): module-mode files compile full-content envs

**The regression** (#569, `getFileEnv` honoring the `module` header):
module-mode header processing imports at `OLeanLevel.exported`, and an
exported-level env is not "the same env minus some private names" — it
is **content-trimmed**:

- every imported public theorem is loaded as a body-less **axiom**
  (probe on Init: `Nat.add_comm`, `Char.succ?_eq`,
  `List.length_append` all `axiom, no value`);
- non-exported `_private.*` constants are absent entirely
  (Init: 14,457 → 2,148 at v4.29; 16,278 → 2,328 at v4.33);
- only `@[expose]`d def bodies survive.

Compiling that env axiomizes imported proofs: kernel checks pass
**vacuously**, every address drifts from private-level compiles of the
same constants (catalog, `getCompileEnv`, the `PrimAddrs` pins), and
because the exported view is level-closed, the groundedness scan sees
**0 ungrounded** — the corruption is silent. That is exactly the
`tc-pins` (`_private...Char.succ?_eq._proof_1_8: resolves in env`) and
`tc-accel-diff` (accel≡pure address parity) failures on main.

v4.33 relevance: `ConstantInfo.value?` now defaults
`allowOpaque := false` (theorem values hidden at the API level — the
"theorems effectively opaque" change). ix reads values by constructor
match everywhere (audited: zero bare `value?`/`hasValue` uses), so only
the import level matters — and at the private level, v4.33 theorem
bodies are fully present.

**The fix** — content and scoping are separated:

- `getFileEnvCore` (new; `getFileEnv` keeps its signature): always
  import classic/private — full content — and report the header's
  `module` bit.
- `moduleVisibleNames`: the exported-level **name** set, used for
  scoping only.
- `Ix.EnvScope` (new module): `collectDeps` moves here from
  `Ix.Cli.ValidateCmd` (namespaced, so tests can import it alongside
  `ValidateAux`'s same-named top-level mirror — `ParityEnv` drops its
  inlined copy), plus `defaultConstList`: classic files keep the
  whole-env behavior byte-identically; module-mode files seed from the
  visible surface + the file's own decls and close over the
  full-content env. Referenced foreign `_private.*` proof auxiliaries
  ship (content is mandatory for groundedness, named rows for
  decompile/tc lookups); unreferenced foreign privates stay out — the
  qualified-package isolation TruthMines' `module-header-v1` patch
  wanted, minus the corruption.
- All five CLI default paths (`compile`, `compile-lean`, `validate`,
  `validate-lean`, `ingress`) route through `defaultConstList`.
- `ParityEnv.ensure` is now **fail-closed** (`allowPartial := false`) —
  silent drops were how this regression hid.

**How we validate theorem bodies are really exposed** (kernel-checking
cannot detect axiomization — axioms typecheck vacuously — so the
assertions are kind/content/address-level):

1. `Tests/Ix/MetaEnv` (default `lake test`): for both header modes, the
   env must hold imported `Nat.add_comm` as a `.thmInfo` whose proof
   refs **all resolve in the same env**, with `_private.*` present; and
   `defaultConstList` must scope names without trimming content
   (theorems stay theorems in the scoped list). Measured on the module
   fixture: 51,981 visible seeds → 64,705 after closure < 65,995 full.
2. `ParityEnv` fail-closed: any missing referenced constant now aborts
   the tc fixture compile instead of silently dropping named rows.
3. The tc gates over a freshly built `tc-parity.ixe`: `tc-pins` proves
   `_private.*` names resolve; `tc-accel-diff` proves address parity
   (addresses are transitive content hashes — they only converge on
   real proof bodies); `tc-roundtrip` proves theorems survive to and
   from the artifact as theorems.

## Commit 3 — fix(bench): lean4lean replay walks theorem proofs again

`Benchmarks/Lean4Lean.lean` vendors the replay machinery, and its
`getUsedConstants` used **bare `ConstantInfo.value?`** — v4.33's new
`allowOpaque := false` default made theorem values invisible to it, so
proof-only auxiliaries (the `._f` structural-recursion helpers, e.g.
`Nat.add_comm._f`) were never walked and the kernel rejected every
replayed theorem depending on one ("unknown constant"). `constHasStrLit`
had the same blind spot. Both now pass `allowOpaque := true`, mirroring
the fix the lean4lean fork itself ships in `Lean4Lean.Replay` at pin
`4844eda4`. (The `lean4lean` ignored gate never ran on the update
branch, so this only surfaced in this PR's gate sweep. Repo-wide audit:
these were the only two bare `value?` uses left — `Ix/` and `Tests/`
read values by constructor match.)

## Verification

| gate | result |
|---|---|
| `cargo check --workspace` | clean |
| `cargo fmt --check` | clean |
| `lake build` (all default targets) | green |
| `lake test` (default suites) | **2,661 assertions, exit 0** |
| `cargo test --workspace --release` | **1,254 passed, 0 failed** |
| tc gates (fresh fail-closed `tc-parity.ixe`, 6,254 consts) | **green**: `tc-pins` 8/8 — incl. `_private...Char.succ?_eq._proof_1_8` in 773ms, the exact red on main; `tc-accel-diff` 5/5 — the exact `BitVec.toNat`/`Nat.decLe` addr-parity reds; `tc-roundtrip` whole-env (165,970 meta-checked); `tc-tutorial` 274; `tc-init`; `tc-anon-diff` 6/6 |
| `lake test --wfail -- --ignored lean4lean` | **green** (37-declaration closure replay; red before commit 3) |
| `lake lint -- --wfail` | green |
| `lake test -- cli` | green |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `nix build .#checks.x86_64-linux.clippy` (the CI leg) | green |
| meta-env suite (the new regression tests) | **3/3 green** |
| catalog / catalog-qualified / import-ixe | green |

One toolchain re-baseline surfaced by the fail-closed parity compile:
v4.33 moved `Std.DTreeMap.Internal.Impl.minKey!_eq_get!_minKey?` to
`Std.Internal.List.minKey!_eq_get!_minKey?` (unique rename, same
statement shape → same OOM-class pin coverage); the seed fixture and
`Pins.lean` label were updated in the merge commit.

## Performance impact

**Module fix: zero change for classic files.** Classic-header files
(every Benchmarks compile file, the mathlib path) short-circuit before
any new code runs — constant lists and artifacts are bit-for-bit
unchanged. Module-mode files' CLI default paths pay two new costs,
measured at real `Ix.lean` scale (239k-const closure):

| added cost | measured |
|---|---|
| second (exported-level) header import | **+1.7 s** (the classic import itself is 1.9 s; native `importModules`, transfers directly to the CLI) |
| scoping closure walk (`defaultConstList`) | 6.2 s interpreted ⇒ **~1–2 s compiled** (same `collectDeps` shape as routine `--consts`/`--ns` runs) |

Compile-input size for a module-mode file: the default scope is
**201,595** consts — 16% *smaller* than the pre-#569 classic full dump
(239,424) and 24% larger than the trimmed 162,329 that current main
compiles. The `.ixe` grows vs main because theorem proofs are actually
stored now — that is the artifact ceasing to be axioms, not overhead.
The lean4lean walker fix likewise just restores 4.29-equivalent work
(the "cheaper" broken walker skipped the proofs the replay then choked
on). `ParityEnv`'s fail-closed flip costs nothing (same compile, 1.36 s
env-ready for 6,254 consts).

**Toolchain bump (inherited).** Core grew ~9% (Init at private level:
60,598 → 65,994 consts), so envs, imports, and compiles scale up
proportionally. No kernel performance-class regressions surfaced: the
tc pins — which exist as OOM/∞/perf-class guards — pass in 2–773 ms
against 60 s budgets, and the whole-env kernel roundtrip peaked at
34.3 GiB, in line with prior same-scale runs. Two caveats: the
inherited `ixvm` ignored-suite OOM (42–79 GiB in the Aiur toplevel
build, HANDOFF §2.3) is the one genuine open perf problem, pre-existing
and untouched here; and the canonical mathlib-scale numbers (69 s Rust
compile, 3.16 GB artifact, validate RSS profiles) have **no v4.33
counterpart yet** — Benchmarks pins a 4.29 mathlib, so A7-scale perf is
unmeasurable until a 4.33-compatible mathlib exists. When it does,
expect wholesale content-address churn and cost-model drift (HANDOFF
already flags the repinned Slice FFT cost as stale): re-baseline,
don't compare byte-for-byte.

## Downstream

- Content addresses churn wholesale vs v4.29 artifacts (toolchain bump:
  derived instances flipped `def`→`theorem` in v4.31, `Float.Model`,
  exposure changes, …) — re-baseline, don't debug (HANDOFF §3).
- TruthMines: re-pin onto this merge; their module-header isolation now
  comes from seed scoping (`defaultConstList`) instead of content
  trimming, so their qualified catalogs keep working while compiled
  artifacts carry real proofs. Their closure-scoped compiles
  (`--consts`) can now resolve `_private.*` spellings again.


samuelburnham:
!benchmark compile decompile

samuelburnham:
!benchmark ooc

samuelburnham:
!benchmark aiur

samuelburnham:
!benchmark aiur fresh

samuelburnham:
!bencher-thresholds-reset all

samuelburnham:
!benchmark compile decompile

arthurpaulino (approved):

PR #576, opened 2026-08-20
ci: Move ignored tests to run on `merge_group`

samuelburnham:
Replaces the ignored test suite on push to main with the same tests in a PR-gating merge queue workflow, such that we can de-queue a failing PR and fix it in place rather than having to triage a merge with errors after the fact. It uses a matrix job to fail fast and save on CI cost.

Also fixes the Valgrind job which doesn't work on binaries compiled with `target-cpu=native`, and re-enables the `-D warnings` rust flag as the default.

Requires the merge queue to be enabled with branch protection settings


.github/workflows/ci.yml
```diff
@@ -88,8 +88,6 @@ jobs:
           use-github-cache: false
       - name: Test Ix CLI
         run: lake test --wfail -- cli
-      - name: IxVM kernel tests
```
samuelburnham, line 91: We could add this back for early signal if helpful, it takes ~2 min

arthurpaulino (dismissed):

arthurpaulino (dismissed):

arthurpaulino (dismissed):

arthurpaulino (approved):

PR #579, opened 2026-08-20
 perf: decompile Pass-2 restructure + compile hot-path batch (post-#572 perf pass)

johnchandlerburnham:
Performance pass over the decompile and compile pipelines, following up on the post-#572 benchmark reports (FLT decompile +114% time / +110% RAM was the headline regression). Three commits on main (post-#578): thin LTO, Rust-side phase/RSS instrumentation, and the algorithmic batch. **Serialized bytes are unchanged everywhere** — verified end-to-end at Mathlib scale (see Validation).

## Headline numbers (Mathlib, 736,624 consts, 32-core / 124 GiB box)

**Decompile** (`ix decompile compilemathlib.ixe`, CI-comparable: local baseline 151.2 s ≈ CI's 152.5 s row):

| | before | after | Δ |
|---|---|---|---|
| Pass 2 ingress | 18.8 s | **6.0 s** | −68 % |
| Pass 2 gen | 75.9 s | **62.2 s** | −18 % |
| **wall total** | **151.2 s** | **120.9 s** | **−20 %** |
| peak RSS | 34.8 GiB | **30.5 GiB** | −4.3 GiB |
| constrained run (65536-entry clear knob, 7 clears) | 228.0 s | **136.1 s** | −40 % |
| cost per kenv clear | ~10 s | ~1.9 s | −81 % |

The constrained row is the FLT mechanism reproduced locally: FLT's dependency closure crosses the old default kenv-clear threshold, so its CI row was paying a full closure re-walk on every clear. That cliff is now flattened three ways (a clear-surviving refs memo + cheap re-ensure + a threshold FLT never reaches), so the FLT row should recover most of the regression **plus** RAM.

**Compile** (`ix compile CompileMathlib.lean`, A/B on this branch's base vs tip, byte-identical outputs):

| | base | perf | Δ |
|---|---|---|---|
| setup_scan | 11.1 s | 9.2 s | −17 % |
| scheduler | 32.8 s | 30.0 s | −8.5 % |
| **compile+serialize+write** | **59.5 s** | **55.0 s** | **−7.6 %** |

### Decompile Pass 2

- **`get_constant_info_references` rewritten** as a single-accumulator iterative DAG walk (visited set keyed on `Expr`'s digest-backed `Hash`/`Eq`) instead of materializing+merging+cloning a `NameSet` per expression node. Feeds both compile's `setup_scan` and decompile's Pass-2 ingress BFS. Result sets are value-identical; no consumer reads set iteration order into bytes (SCC members and every serialized section re-sort canonically).
- **Ingress BFS refs memo survives kenv clears**: the BFS's `ingressed` set doubles as the kenv-content tracker, so the periodic kenv clear wiped the only BFS memo and every clear re-walked the whole closure — a Θ(N²/L) term, and the mechanism behind FLT's +114 %. The discovered reference lists now survive clears (the reference graph doesn't change when the kenv does); names absent from `work_env` stay unmemoized so late-appearing aux constants are re-probed live.
- **Aux ingress walker dedup moves into `KernelCtx`**: `ingress_target_type_deps` / `ingress_field_deps` kept a per-call `seen`, re-expanding the full reference closure for every block (Θ(blocks × closure)). The set is now kenv-lifetime (`aux_ingress_seen`), keyed by **resolved id** — not bare name — so an address shifted by later aux registration reads as unseen and re-ingresses under the new id, exactly as the per-call sets healed it. Fresh-per-block in compile (§10.5 unchanged); cleared with the kenv in decompile. `collect_const_refs` additionally gains a digest-keyed visited set (shared subterms walked once, not once per occurrence). Measured: Pass-2 gen −18 %.
- **Pre-warm ingests proofs as type-only stubs**: the Pass-2 BFS ingested theorem/opaque constants with full proof terms (~47 KB/name measured) while the gen-phase walkers deliberately use type-only stubs for the same kinds. The new `ensure_in_kenv_of_prewarm` ingests thm/opaque as type-only `Axio` stubs — sufficient for kernel TC (proof irrelevance; opaque never unfolds) and the exact fidelity `ingress_type_stub` already gives. With per-entry kenv cost collapsed, the clear default rises 131072 → 1048576 and becomes a rare backstop; the `IX_DECOMPILE_KENV_CLEAR_ENTRIES` knob semantics are unchanged. Validated by a full-Mathlib decompile with **0 aux_gen errors** across 6,598 regenerated blocks (plus the congruence checks).
- Free wins: `lean_env` (a third whole-env copy) drops after Pass 1.5 instead of living to end of `decompile_env`; `rs_decompile_env` skips the ~30 GiB destructor churn on exit (one-shot CLI, mirroring `rs_compile_env`) and reports read/parse spans — the pre-Pass-1 `.ixe` parse is ~37 s of the wall and is now visible (largest remaining chunk; future target).

Lean mirrors land in the same commit: `AuxKernelCtx.auxIngressSeen` + `drainIngressQueue` + `collectConstRefs` visited set + `ensureInKenvOfPrewarm` wired into both Pass-2 driver variants (`Ix/AuxGen/Kernel.lean`, `Ix/AuxGen/Recursor.lean`, `Ix/DecompileDriver.lean`).

### Compile

- **Digest-keyed maps drop SipHash**: every blake3-digest-keyed map in `ixon/sharing.rs` moves from `std::HashMap` to `FxHashMap` (keys are already uniform digests; every order-sensitive consumer re-sorts).
- **Merkle root computed once, in parallel**: new host-only `merkle_root_canonical_sorted` (parallel leaf/level hashing for callers holding sorted unique leaves, bit-identical root). `rs_compile_env` computes the const table + root **once** and hands both to the new `put_file_with_header`; previously the par-sort and the ~2·N-node tree ran twice per compile.
- **Block-context addresses hoisted**: `compile_definition/inductive/recursor` take the block-context address list as a parameter, computed once per block (was: full ctx sort + re-intern per member, O(M² log M) per block).
- **Metadata moves instead of cloning**: mutual-path `ConstantMeta` moves out of `all_metas` via `remove` instead of two deep clones per constant.
- **Instrumentation**: `IX_COMPILE_DBG=1` now lights the Rust side too (scheduler phases with RSS attribution, `Env::put_file` per-section timers — the §4 serial name topo-sort is timed separately and shows 1.9 s at Mathlib scale, the next serializer target).
- **Thin LTO** on the release profile (no embedded guest builds in this workspace; external guest builds use their own profiles).

## Validation

- **Whole-Mathlib byte identity**: `ix compile` of `CompileMathlib.lean` on this branch's pre-perf base vs tip → `cmp` **identical** (3,326,436,547 bytes).
- **Whole-Mathlib decompile**: 0 aux_gen errors, 0 congruence failures (default and constrained-knob runs).
- `--ignored` parity gates: **validate-aux** (3,699 matches / 0 mismatches), **aux-gen-diff** (plain-defn mismatches 0), **decompile-diff** (plain fidelity + callsite replay PASS; Pass-2 plan parity 300/300; aux-fidelity PASS) — these cover the Lean mirror changes.
- `cargo test --workspace --release`: 1,256 passed / 0 failed (incl. `put_file_matches_put` for the `put_file_with_header` split).
- `lake test`: 2,668 checks, exit 0.
- `cargo clippy --workspace --release --all-targets` clean; `cargo fmt --check` clean.

## Explicitly not touched

Fresh-per-block `KernelCtx` in the compile scheduler (§10.5 determinism), wire format (no `Env::VERSION` change), DEMOTE/EAGER defaults, and the Lean drivers' performance (mirrors are behavior-parity only, per BENCHMARKS.md ground rules).

Follow-up targets surfaced by the new instrumentation (measured, not yet done): the ~37 s decompile pre-phase (`.ixe` parse — per-member Muts block re-parsing lives there), the §4 name topo-sort in the serializer (1.9 s serial), and the prove-side items from the benchmark digest (targeted hint-row promotion, FFT pin re-baseline, sequencing the two ixvm test envs).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,10 @@ panic = "abort"
 
 [profile.release]
 panic = "abort"
+# Cross-crate inlining across the ixon/compile/ffi boundary — the hot
+# loops (blake3 digests, subterm sharing, serialization) span crates.
+# "thin" keeps build times tolerable vs "fat" for near-equal wins.
+lto = "thin"
 
 [workspace.package]
 version = "0.1.0"
--- a/Ix/AuxGen/Kernel.lean
+++ b/Ix/AuxGen/Kernel.lean
@@ -137,6 +137,15 @@ partial def kunivToLevel (u : MKUniv) (paramNames : Array Name) : Level :=
 structure AuxKernelCtx where
   tcState : Ix.Tc.TcState .meta
   ingressCache : Std.HashMap (Address × Address) MKExpr := {}
+  /-- Mirrors Rust `KernelCtx.aux_ingress_seen`: ids whose
+      `ingressAuxGenDep` dispatch already ran against this kenv. The
+      dispatch is deterministic per constant kind, so a seen id is at
+      final kenv fidelity and the ingress walkers skip re-expanding its
+      reference closure. Keyed by resolved id (not bare name) so an
+      address shifted by later aux registration reads as unseen. Lives
+      as long as the kenv (this ctx never clears — see the Pass-2
+      driver's KENV_CLEAR note). -/
+  auxIngressSeen : Std.HashSet MKId := {}
 
 /-- Fresh bridge context: empty Meta kenv, canonical prim addresses with
     no resolver (prim KIds get synthetic display names; the accelerations
@@ -343,7 +352,7 @@ def ensurePreludeInKenvOf (maps : AddrMaps) : KBridgeM Unit := do
     deps surface as `unknownConst` at TC time and are faulted in).
     Mirrors Rust `ensure_in_kenv_of_inner_env` (expr_utils.rs:1944). -/
 partial def ensureInKenvOfInner (name : Name) (maps : AddrMaps)
-    (replaceAxioStub : Bool) : KBridgeM Unit := do
+    (replaceAxioStub : Bool) (stubProofValues : Bool) : KBridgeM Unit := do
   let addr := maps.resolve name
   let zid : MKId := ⟨addr, name⟩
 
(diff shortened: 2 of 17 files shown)


johnchandlerburnham:
!benchmark compile decompile

johnchandlerburnham:
!benchmark ooc

johnchandlerburnham:
!benchmark aiur

Ix/Aiur/Statistics.lean
```diff
@@ -97,6 +106,8 @@ def computeStats (compiled : CompiledToplevel) (queryCounts : Array QueryCount)
   let mkStats (name : String) (shape : CircuitShape) (h hits : Nat) : CircuitStats :=
     { name, width := shape.committedWidth, height := h, cacheHits := hits,
       fftCost := fftCost shape h logBlowup,
+      paddedFftCost :=
+        if h == 0 then 0.0 else fftCost shape h.nextPowerOfTwo logBlowup,
```
arthurpaulino: I'm not sure if this is the right formula.
I think trace heights are padded, not the full FFT cost.
@gabriel-barrett knows better.

samuelburnham (approved):

PR #583, opened 2026-08-20
perf(ixvm): substitution sharing, Int primitives, projection def-eq; struct-eta scope fix

arthurpaulino:
Five commits on `ap/ixvm`: three kernel performance changes, one WHNF soundness/robustness fix, and the codegen regeneration with repinned FFT costs.

### perf(ixvm): share argument-independent substitution work
(`Ix/IxVM/Kernel/Subst.lean`)

Large beta-reduction traces instantiate the same expression with many different arguments. Since the argument is part of the memo key, subtrees that never mention the substituted binder were rebuilt once per argument.

- New `expr_has_bvar_at(e, depth)` predicate (with binder/let helpers) detects whether the *exact* binder occurs, distinguishing "mentions the binder" from "mentions only variables above it".
- `expr_inst1`: when the binder does not occur but higher loose bvars do, route through `expr_lower(e, 1, depth + 1)` — a memoized, argument-independent path shared by all arguments.
- `expr_inst_many`: analogous `has_bvar_in_range(e, depth, depth + n)` window check, routing binder-free subtrees through `expr_lower(e, n, depth + n)`, plus an explicit empty-substitution-list fast path.
- Ordinary substitution walks and the `expr_lower_walk` circuit layout are unchanged.

Validation: focused bytecode check of `Algebra.TensorProduct.toLinearEquiv_tensorTensorTensorComm` passes at 403.031B FFT.

### fix(ixvm): decline struct eta for out-of-scope majors
(`Ix/IxVM/Kernel/Whnf.lean`)

Struct-eta iota is speculative, but the code unconditionally ran `k_infer` on the stuck major. During canonical recursor reconstruction, WHNF can hit a loose major under a deliberately empty local type context; inference then indexed past `types` and aborted on an unmatched list case.

- Guard added before inference: `memo_u32_less_than(list_length(types), expr_lbr(major))`. If the major's loose-bvar range exceeds the supplied context, decline eta (return stuck) and let ordinary reduction continue — matching reference behavior when speculative inference cannot establish a type.

Validation: fixes `SSet.prodStdSimplex.pairingCore.Type₁.rec`, the shared failure in FLT shard 145 and Mathlib shard 226; focused constant and both shards pass under bytecode interpretation.

### perf(ixvm): reduce closed Int primitives natively
(`Ix/IxVM/Kernel/NatPrim.lean`, `Ix/IxVM/Kernel/Whnf.lean`)

Closed Int arithmetic was previously reduced by unfolding the logical Int definitions into large Nat reduction chains during WHNF.

- New primitive family `6 = Int` in the address-keyed `prim_family` classification, dispatched via `try_int_prim_dispatch`.
- Hardcoded addresses for `Int.ofNat`, `Int.negSucc`, `add`, `sub`, `mul`, `neg`, `emod`, `ediv`, `bmod`, `bdiv`, `natAbs`, `pow`.
- Canonical Int literals extracted as `(sign, magnitude)` pairs over `KLimbs`; closed arithmetic, Euclidean and balanced division/modulus, `natAbs`, and powers evaluate natively.

Validation: primitive addresses covered by the elaborated IxVM address-parity suite. The focused Batteries Char casing proof still needs further def-eq work, but the catastrophic Int expansion on that path is gone.

### perf(ixvm): compare projection applications structurally
(`Ix/IxVM/Kernel/DefEq.lean`)

Projection-heavy proof terms triggered pathological eager WHNF in the def-eq slow path.

- After bounded lazy reduction (`k_is_def_eq_struct_bounded`, budget 8) fails, try `k_is_def_eq_structure_tree` before `slow2_eager_fallback`.
- `try_same_proj_head_app`: spines with equal length whose heads are projections with the same struct address, same index, and the same inner-expression pointer compare argument-wise only. (Pointer equality used in the positive direction only; mismatch just declines the shortcut.)
- `k_is_def_eq_structure_tree_app`: descend through `App` nodes recursively without resetting the reduction budget.
- Unrelated shapes still fall through to the existing eager fallback.

Validation: complete 600-shard sweep across Init, InitStd, Lean, Batteries, Mathlib, and FLT (24 workers, five-minute shard timeout) — all shards passed.

### test(ixvm): regen codegen and repin FFT costs
(`crates/ixvm-codegen/src/aiur_ixvm.rs`, `Tests/Ix/IxVM.lean`, `Tests/Main.lean`)

- Regenerated `aiur_ixvm.rs` via `ix codegen` (784 IxVM functions).
- Repinned 77 kernel FFT entries and the shard pipeline pin (`6_720_706_901` → `6_770_109_972`) from a full `lake test -- --ignored ixvm` run; rerun passes clean.
- Pins moved up ~0.5–1.5% across the board — the cost of the new Int-primitive classification and def-eq machinery on the pinned suite, traded against the large wins on the pathological constants above.


arthurpaulino:
!benchmark aiur

samuelburnham (approved):

PR #584, opened 2026-08-20
`ix catalog` at corpus scale

johnchandlerburnham:
Makes `ix catalog` viable for its first real consumer: a Lean corpus of ~119 pinned packages (~81 building as one union, ~40 of them sharing a mathlib closure) whose entire build becomes one `ix catalog` invocation. Per the plan: one memory blocker fixed, one latent correctness hole gated, a fail-closed coverage diagnostic, spec-file/audit/report ergonomics, and the in-repo benchmark that both catches the blocker and — as it turned out — found four more scaling walls.

## The series

1. **catalog: pin loader imports to `OLeanLevel.private` + level-regression gates** (Item 4 / I4)
   The catalog depended on `importModules`' *default* level for D6 "complete bodies". An exported-level env axiomizes imported proofs and drops `_private.*`; kernel replay passes vacuously and `--audit` compares two identically-axiomized legs (#572 documents this shipping in #569) — no existing gate could catch a regression. Now explicit at both call sites, gated by an always-run base-env check and a loader-level fixtures test.

2. **catalog: fail closed on un-replayed cross-package roots** (Item 5 / I5)
   A member's closure may reach a provider module the provider's own roots don't cover; its constants were renamed but never replayed → bare kernel `unknown constant`. Now a pre-replay coverage fold rejects with the member, module, owner, and the fix ("Add M to Q's roots"); misordered members are caught the same way.

3. **catalog: stream member environments — load, stage, free, one at a time** (Item 1 / I1, the blocker)
   `resolveLibs` held every member env live; at corpus scale that is ~40× a mathlib env. Streaming requires actually *freeing* compacted olean regions, which requires staged declarations to share nothing with them — a fused rename + structural copy. `forEachLib` streams members; the toolchain base imports once after the pass; staged declarations kernel-replay afterwards. `auditCatalog` streams the same way, and its address comparison doubles as a copy-fidelity gate.
   Two Lean-runtime sharp edges found as segfaults and now permanently guarded: the code generator **eta-reduces same-fields constructor rebuilds to the original (region-resident!) object** (`.bvar`, `SourceInfo.synthetic`, `DataValue.ofBool` — even through an `if`/`else` rebuild — `ReducibilityHints.regular`), and `{ val with … }` record updates silently share boxed fields. All-scalar rebuilds route one field through arithmetic the simplifier can't fold, and the copier self-checks with `ptrAddrUnsafe` panics — a future codegen change fails loudly at staging time instead of corrupting kernel replay.

4. **bench: refresh CatalogReal member pins for Lean v4.33.0**
   The ignored benchmark had been unbuildable since the v4.33 consolidation (pre-4.33 batteries rev). New pins = the mathlib-v4.33.0-inherited revs (same as `Benchmarks/Compile` and the downstream corpus). The #569-era baselines belong to the old pins; refreshed baselines below.

5. **catalog: harden the region-evicting copier at real-library scale**
   Two more pass-through paths found at CatalogReal scale: `Array.mapM`'s unsafe implementation **returns the input array object when it is empty** (empty `Syntax.node` args in `ofSyntax` mdata; Lists are immune — `[]` is a tagged scalar), and the remaining copiers gained the same self-check panics.

6. **catalog: `--spec` file input** (Item 3 / I3)
   JSON spec file, mutually exclusive with positionals; `groups` reserved for Item 2; parsed in the library (unit-tested — CLI modules only build under the exe build); resolved spec + path echoed into `--report`. Byte-identical to the positional build on CatalogReal.

7. **catalog: `--audit-only` subset gating + peak RSS in `--report`** (Items 6+7 / I6, I7)
   `--audit-only Qual[,…]` runs the §3.1 invariant on a rotating subset while still writing the artifact (the full audit is N+1 compiles — a multi-hour session at corpus scale). `--report` gains `peakRssBytes` (process-tree VmHWM via the existing texray sampler) and `auditedQualifiers`.

8. **catalog: hash-cons the region-evicting copy**
   First mathlib-spine wall: olean compaction dedups only *within* a module, so cross-module pointer sharing is ~zero (measured pointer-cache hit rate over batteries: 16%) and a purely pointer-cached copy materializes the member's entire syntactic content. The copy is now hash-consed — children intern before parents, so structural equality is shallow and the cached hash fields make it O(1) — and CatalogReal replay got *faster* (kernel sharecommon receives pre-shared input).

9. **catalog: bound staging at mathlib scale — per-decl caches, copy/share hybrid, unconditional replay**
   Three more walls from the spine leg (Mathlib member: 569,200 owned constants): source-pointer caches grew with total *visited* content (~10⁹ nodes; the process hit `vm.max_map_count`) → now reset per declaration; even hash-consed, heap-materializing a mathlib-scale member measures ~10× its compacted form → members owning > `copyStageMaxOwned` (default 100k) constants are staged **sharing their env's regions, which stay mapped** (`EnvDisposal.keepRegions`) — the corpus shape is exactly the favorable case (many small-content members with heavy closures: copy + free; a handful of heavy-content members: keep + share); and replay passed empty `Options` to `addDecl`, so default `maxHeartbeats` rejected mathlib's heavy proofs with a deterministic timeout (latent since #569) → replay now runs unlimited — the declarations were already accepted once.

10. **bench: mathlib-spine catalog with a peak-RSS gate** (Item 8 / I8)
    `Benchmarks/CatalogSpine`: mathlib's dependency spine + Mathlib + three small real-corpus dependents (BET, GibbsMeasure, KolmogorovExtension4) whose envs each contain the full mathlib closure. `lake test -- --ignored catalog-spine` runs baseline and spine legs as `ix` subprocesses via `--spec`/`--report` (dogfooding I3/I7) and asserts the spine peak ≤ 1.6× baseline. **Green.**

## Measurements (124 GiB box, 32 cores, v4.33 pins)

| Run | Result |
|---|---|
| CatalogReal build (streaming) | 20,117 owned; 18,715 replayed; 146,987-const union ~16 s; 360,552,416 B at root `fb0640061a61…` |
| CatalogReal `--audit` | 20,117/20,117 verified in 121 s |
| `ix check-rs realcat.ixe` | 146,987/146,987 pass (19.3 s) |
| pre-streaming vs streaming artifact | **byte-identical** (`cmp` + `ix diff` clean, same root) |
| `--spec` vs positional artifact | **byte-identical** |
| copy-staged vs region-shared staging | identical roots (fixtures + CatalogReal byte-identity) |
| Spine baseline leg (9 members, 590,748 owned) | 580,510 decls replayed → 774,528-const union in ~17 min; 3,332,486,766 B at root `6e8bd09c14c6…`; **peak 32.2 GiB** (was: 119 GiB OOM panic pre-hybrid; unbounded pre-streaming) |
| Spine full leg (12 members) | 580,927 decls → 774,963-const union; 3,334,313,087 B at root `3825539457d6…`; **peak 35.3 GiB — ratio 1.10×** |
| `ix check-rs` on the spine artifact | 774,962/774,963 at default fuel — `Spine.BET.periodicPt_mem_recurrentSet` exhausts the Rust kernel's `MAX_REC_FUEL` (10M); with `IX_MAX_REC_FUEL=1000000000`, **774,963/774,963 pass** (399 s). Pre-existing checker ceiling, not a catalog defect (the Lean kernel re-accepted the declaration during replay) |
| I8 gate (spine ≤ 1.6× baseline) | **passes** |

Trade-off note: at CatalogReal scale (three small members, overlapping closures) streaming costs more than the old hold-everything loader (5.9 → 10.6 GiB build-only peak) — the copy-out materializes union content the old code aliased from regions. The win is boundedness where it matters: N members sharing a heavy closure.

## Gates

`lake test -- catalog` / `catalog-qualified` / `import-ixe` / `--ignored catalog-fixtures` (now also I4/I5/I6/hybrid tests) / `--ignored catalog-spine`, `lake build ix` — all green, run unpiped.

## Non-goals

Per the plan: no olean/source emission, no instance/attribute/LCNF transfer, no downstream corpus records or nix packaging, no `import_ixe` materialization changes. Item 2 (grouped loading) deferred: streaming + the hybrid bring the spine into range; grouping remains available if corpus-scale import wall-clock (measured downstream) demands it.


arthurpaulino (approved):
This can probably help us in the future: https://github.com/toon-format/toon

PR #585, opened 2026-08-21
perf: lazy §5 metadata windows on the demoted env load path

johnchandlerburnham:
## Headline numbers (Mathlib, 771,135 consts / 3.33 GB env, A/B on the same box)

`ix decompile`, section split from the new `[Env::get]` timers:

| | before | after | Δ |
|---|---|---|---|
| §5 named metadata parse | 12.02 s | **3.19 s** | −73 % |
| §2 consts (blake3 verify) | 1.56 s | 0.95 s | −39 % |
| merkle root recompute | 0.15 s | 0.01 s | −93 % |
| **whole `.ixe` parse** | **16.56 s** | **7.17 s** | **−57 %** |
| **decompile wall** | **107.1 s** | **97.1 s** | **−9.4 %** |
| peak RSS | 31.6 GiB | 30.3 GiB | −1.3 GiB |

Every pass timing is flat (muts plan 0.60→0.56 s, Pass 1 14.6→14.5 s, Pass 2 65.1→64.9 s): the work removed from the load was not pushed into the passes.

The same loader serves `import_ixe`, so a selective import (the TruthMines process-boundary use) stops paying a full §5 metadata parse for the whole env before materializing its first constant.

## What changed

- **Lazy §5 metadata windows** (the win): `Env::get_demoted_named` used to parse every §5 entry's metadata into the structured form and immediately re-serialize it into the demoted bytes repr — paying a full parse *plus* a full re-encode per entry, serially, before any pass starts. It now keeps each entry's `meta_len`-framed window verbatim (one contiguous arena plus the shared §4 reverse index) and decodes on demand: new `MetaRepr::Window` / `LazyMetaWindow` in `env.rs`, decoding through the same three grammar helpers `get_named_indexed` itself now parses with (single source for the window grammar). A `OnceLock` split memo (meta length + original address, ~40 B/entry) is filled by any meta decode — the decompiler's muts-plan sweep reads every entry's meta up front, so `has_original()`/`original()` on the passes stay as cheap as the old materialized slots. The mutating original-slot APIs materialize the window first, so `Named` stays total. RSS drops because the indexed windows are the compact encoding (varint name indices) while the old demoted blobs re-encoded every name reference as a raw 32-byte address.
  - Trade-off, documented on `get_demoted_named`: interior §5 corruption that the eager loaders reject at load now surfaces as a panic on first decode (window *framing* is still validated at load; §2 constants keep their per-entry hash check).
- **§2 integrity check parallelized**: serial framing/order pass first, then the per-const blake3 verification sweeps in parallel (serial on the guest target, same cfg shape as the existing merkle pair).
- **Read-side merkle root**: `Env::get` recomputed the root with the serial builder even though §2's order check proves the key set sorted and duplicate-free — now uses the parallel `merkle_root_canonical_sorted` from #579 (host-only, same cfg pattern as the write side).
- **§4 name-table topo sort** (write path): the recursive DFS is now an iterative ancestor-chain walk with a pre-sized visited set. The emission order — which is wire bytes — is reproduced exactly; proven by the whole-Mathlib byte compare.
- **`[Env::get]` per-section timers** behind `IX_VERBOSE`/`IX_COMPILE_DBG`, mirroring `put_file`'s — these produced the section split above.
- §4 names-lookup pre-sizing (it rehash-doubled through 4.7M inserts; measured a wash at Mathlib scale, kept as principled).

## Validation

- **Whole-Mathlib byte identity** (write path): `ix compile` of `CompileMathlib.lean` at this branch vs the pre-branch reference → `cmp` **identical** (3,326,436,547 bytes).
- **Whole-Mathlib `ix validate`**: 0 failures across all phases (compile, aux congruence, two decompiles through the lazy loader, serialize, per-constant roundtrip fidelity).
- **New test** `demoted_load_matches_structured_load`: lazy load ≡ structured load on every `Named` accessor (meta, original, has_original, hints), mutation APIs materialize correctly, and the two loads re-serialize byte-identically — including entries whose metadata carries name references through the §4 index. (It also surfaced a pre-existing latent quirk, unchanged here: §4 emits the parent-closure of `env.names`, so a *synthetic* env whose multi-component names lack registered parents permutes §4 on reload. Pipeline envs register every component — whole-env byte roundtrips hold in CI.)
- `--ignored` gates: **validate-aux** 0 failures; **aux-gen-diff** drift/patches/plans/driver gates PASS with all gated mismatch counts 0; **decompile-diff** plain 3,660/0 and aux-family 2,504/0, Pass-2 plan parity 300/300.
- `cargo test --workspace --release`: all suites pass (0 failures).
- Both CI clippy variants (`--all-features` dev and `--release`) with `-D warnings`, and `cargo fmt --check`: clean.

## Explicitly not touched

Wire format (no `Env::VERSION` change), `Env::get`'s eager/validating semantics, the Lean-side deserializers, and the decompile passes themselves. Remaining decompile targets after this: Pass 2 gen (59 s, the dominant phase — per-access `LazyConstant` re-parse territory) and Pass 1 (14.5 s), plus the prove-side items tracked separately.


johnchandlerburnham:
!benchmark compile decompile

arthurpaulino (approved):

PR #589, opened 2026-08-22
feat: Parallel Aiur execution over env shards

samuelburnham:
# Parallel Aiur execution over separate records; RAM-gated shard batches

Whole-env Aiur (IxVM) checking now runs as **one rayon batch over the
partition's shards** — parallel, not concurrent: each shard executes
through the unchanged single-shard machinery into its own private
`QueryRecord`, sharing only the read-only env and toplevel. Peak RSS is
bounded by a **byte-weighted RAM gate** (a std Mutex+Condvar semaphore:
each shard reserves its estimated execution RSS against 85% of available
RAM before starting), so the pool runs at full core width with no
`--jobs` tuning on any box/env pairing. The Lean-task scheduler this
replaces (chunks of N with a full barrier per chunk) measured 2.3–2.5×
slower and is deleted, along with the `--batch` flag — the rayon batch
is now the only whole-partition path.

Also in this PR:

- **Per-claim parallel batch** (`ix check --ixe E names… --jobs N`):
  rayon over resolved addresses for name-list workflows.
- **Analytic prover-RAM projections**: every executed shard reports its
  projected prover peak (three-phase model, calibrated to +1.6–1.7% on
  real ~472 GiB proves) — the input to split/merge repartitioning.
- **Naive `--max-ram` sizing** for the static shard strategy: shard
  count from serialized env bytes × measured ~20,000× amplification,
  targeting ⅔ budget. Median measured projections landed within a few
  percent of target on every env below.
- **`aiur-sharded-env` CI benchmark**: whole-env execution over ISLB
  (new Init+Std+Lean+Batteries env), FLT, and Mathlib — runner-sized
  manifests (`--max-ram 100`), one gated full-width batch per env,
  self-reported bencher rows (check-time, throughput, peak-rss,
  constants, shards).

## Gated full-width sweep (no `--jobs`, naive `--max-ram 400` manifests)

Xeon 6975P-C, 32 cores / 495 GB. Every shard of every env passes; the
gate held peak RSS without a hand-tuned width anywhere.

| env | `.ixe` | shards | wall | peak RSS |
|---|---:|---:|---:|---:|
| init | 194 MB | 14 | 51.8 s | 149 GB |
| initstd | 340 MB | 24 | 1:10 | 255 GB |
| batteries | 326 MB | 24 | 1:18 | 251 GB |
| lean | 504 MB | 36 | 1:40 | 330 GB |
| mathlib | 3.33 GB | 233 | 13:38 | 425 GB |
| flt | 3.44 GB | 241 | 13:22 | 420 GB |

Reference points: FLT at a hand-tuned `--jobs 16` took 18:16 (the gate's
full width is 27% faster); the same 241 shards as 241 processes under
`xargs -P16` took 28:34 — the in-process batch pays the env parse once
where multi-process pays it per shard, a 1.4–1.6× tax at env scale.

### DIFF
--- a/Ix/Aiur/Semantics/BytecodeFfi.lean
+++ b/Ix/Aiur/Semantics/BytecodeFfi.lean
@@ -196,6 +196,26 @@ def checkAddrWithEnv (toplevel : @& Bytecode.Toplevel)
   (checkAddrWithEnv' toplevel funIdx envHandle addrBytes useBytecode).map
     fun r => (r.output, .ofArrays r.ioData r.ioMap, r.queryCounts)
 
+@[extern "rs_aiur_toplevel_check_addrs_with_env"]
+private opaque checkAddrsWithEnv' : @& Bytecode.Toplevel →
+  @& Bytecode.FunIdx → @& EnvHandle → @& ByteArray → Bool → @& Nat →
+    Except String (Array (String × String))
+
+/-- Batch PARALLEL check of many full-closure claims: `addrsBlob` is
+    the concatenation of 32-byte addresses, and Rust checks them on a
+    rayon pool of `jobs` threads (0 = rayon default) — each claim
+    through the exact single-claim machinery over task-private data
+    (own witness io, own query record), nothing shared between tasks
+    but the read-only toplevel and env. Returns the FAILURES as
+    `(addrHex, error)` pairs; empty means all passed. Per-claim
+    outputs/records are not returned — use `checkAddrWithEnv` for a
+    single claim's full result. -/
+def checkAddrsWithEnv (toplevel : @& Bytecode.Toplevel)
+  (funIdx : @& Bytecode.FunIdx) (envHandle : @& EnvHandle)
+  (addrsBlob : ByteArray) (useBytecode : Bool := false) (jobs : Nat := 0)
+  : Except String (Array (String × String)) :=
+  checkAddrsWithEnv' toplevel funIdx envHandle addrsBlob useBytecode jobs
+
 @[extern "rs_aiur_toplevel_shard_check_with_env"]
 private opaque shardCheckWithEnv' : @& Bytecode.Toplevel →
   @& Bytecode.FunIdx → @& EnvHandle → @& ByteArray → Bool →
--- a/Ix/Aiur/Statistics.lean
+++ b/Ix/Aiur/Statistics.lean
@@ -53,6 +53,19 @@ structure ExecutionStats where
   totalCacheHits : Nat
   deriving Inhabited
 
+/-- Coarse PROVER RAM projection for a record with these circuit
+    heights: the committed trace bytes over the LDE domain,
+    `Σ nextPowerOfTwo(height) · committedWidth · 8 · 2^logBlowup`, with empty
+    circuits contributing nothing (the prover deactivates them). A
+    surrogate for relative shard sizing — how many shards must split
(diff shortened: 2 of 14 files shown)


Ix/Aiur/Semantics/BytecodeFfi.lean
```diff
@@ -196,6 +196,26 @@ def checkAddrWithEnv (toplevel : @& Bytecode.Toplevel)
...
+    (own witness io, own query record), nothing shared between tasks
+    but the read-only toplevel and env. Returns the FAILURES as
+    `(addrHex, error)` pairs; empty means all passed. Per-claim
+    outputs/records are not returned — use `checkAddrWithEnv` for a
+    single claim's full result. -/
+def checkAddrsWithEnv (toplevel : @& Bytecode.Toplevel)
+  (funIdx : @& Bytecode.FunIdx) (envHandle : @& EnvHandle)
+  (addrsBlob : ByteArray) (useBytecode : Bool := false) (jobs : Nat := 0)
+  : Except String (Array (String × String)) :=
+  checkAddrsWithEnv' toplevel funIdx envHandle addrsBlob useBytecode jobs
```
arthurpaulino, line 217: This is a bit unfortunate, but this function, along with `shardCheckWithEnv'`, shouldn't be here.
The scope of `Ix/Aiur` folder should be agnostic to Aiur use cases (such as `ix`).

I'm commenting here as a note for myself, because I think I started this mess.

Ix/Aiur/Statistics.lean
```diff
@@ -53,6 +53,25 @@ structure ExecutionStats where
   totalCacheHits : Nat
   deriving Inhabited
 
+/-- Least power of two ≥ `n` (and ≥ 1). -/
+private def nextPow2 (n : Nat) : Nat := Id.run do
```
arthurpaulino: I believe `Nat` already has a `nextPowerOfTwo` or something similar

Ix/Cli/CheckCmd.lean
```diff
@@ -701,7 +857,9 @@ def checkCmd : Cli.Cmd := `[Cli|
...
     "shard"     : Nat;      "0-based shard index K (with --ixe + --ixes): check the constants owned by shard K of the manifest's partition."
-    "jobs"      : Nat;      "Max shards to check concurrently when checking a whole partition (--ixes without --shard). Default: all at once. Lower it to bound peak RAM — each in-flight shard re-ingests its closure into its own IO buffer."
+    "jobs"      : Nat;      "Parallelism. With --ixes (no --shard): max shards checked concurrently (default: all at once). With --ixe alone and N ≠ 1: check the targeted constants on N Rust threads (0 = all cores), each claim over its own private record — peak RAM is bounded by N in-flight claim closures."
+    "json"      : String;   "With --ixes (no --shard): append one env-keyed results row (see Ix.Benchmark.Results) for the batch to this file — check-time, throughput, peak-rss, constants, shards. Used by `ix bench run --backend aiur-sharded-env`."
+    "json-name" : String;   "Row key for the --json row (default: `env`)."
```
arthurpaulino, line 862: The `check` command is getting too complicated. We should strive for an `ix` CLI standardization/simplification revamp soon.

samuelburnham:
!benchmark aiur-sharded-env

arthurpaulino (approved):

PR #591, opened 2026-08-24
ci: Fix benchmarks

samuelburnham:
- Fixes a `!benchmark` comment caching issue caused by bencher API drift, now restores properly from main when `fresh` is not passed as an argument
- Fixes the lean-update action to recursively update the Lean toolchain of all `Benchmarks/**` subdirs 
- Fixes the bencher plot titles for Aiur sharded env check


samuelburnham:
!benchmark ooc

samuelburnham:
!benchmark aiur-sharded-env

arthurpaulino (approved):

PR #594, opened 2026-08-25
formalization: verify Ix.Tc and establish the Ixon v2 compiler refinement path

johnchandlerburnham:
## Summary

On top of the Lean 4.33.1 and `argumentcomputer/lean4ix` baseline already on
`main`, this PR brings together three connected pieces of work:

1. an audited Ix.Tc verification layer for production checking, ingress,
   block transactions, and the currently supported inductive, projection, and
   quotient slices;
2. Ixon v2, which records usage and ownership modes on binders in both Lean
   and Rust while conservatively embedding ordinary Lean terms; and
3. a new compiler-verification library that relates source Lean values,
   production compiler execution, Ixon values and metadata arenas, complete
   expression/constant wire codecs, and Lean4Ix's independent `Theory`
   semantics.

The compiler theorem is intentionally independent of Ix.Tc acceptance. The
checker and compiler now meet at the same Lean4Ix semantic endpoint, but a
compiler theorem cannot prove correctness merely by showing that the output
checker accepts its own encoding.

This is a substantial compiler checkpoint rather than the end of the roadmap.
It proves the ordinary, surgery-free expression path, the complete explicit
wire domain for every expression and `ConstantInfo` constructor, and
codec-safe production compilation for singleton declarations, inductive
families, recursors, sharing, and true mutual blocks. The top-level mutual
entry now constructs collection and bounded classification from source
readiness instead of accepting raw execution witnesses. Declaration-level
semantic preservation, whole environments, canonical-address invariance,
richer mode-producing compilers, and production Rust refinement remain later
milestones.

## Why this is one PR

These changes share one representation and trust boundary:

- Ix.Tc interprets potentially hostile Ixon and must show that successful
  checking admits valid Lean4Ix declarations.
- The compiler starts with kernel-admitted Lean objects and must show that its
  Ixon output has the same independently stated value.
- Ixon v2 is the format both sides exchange, so its binder modes, table
  semantics, metadata, and byte grammar must be represented consistently in
  Lean, Rust, ingress, decompilation, tests, and proofs.

Landing only one layer would either leave the new format without its semantic
bridge or leave the proofs pinned to the obsolete representation. This stack
keeps the format transition, production behavior, semantic relations, and
trust audits reviewable at the same commit boundary.

## Main changes

The base already contains the Lean 4.33.1 toolchain update, current Lean4Ix
pin, TruthMines catalog work, compiler/kernel corpus fixes, independent
TruthMines-library validation, and opt-in CUDA proving. Those merged changes
are not part of this PR's claimed scope.

### 1. Formal-verification library and CI boundaries

- Adds `IxCompileVerify` as a separate, non-default Lake library so compiler
  proofs cannot accidentally use Ix.Tc acceptance as their specification.
- Keeps the large Ix.Tc and compiler proof libraries outside the default
  runtime build while exposing focused build and audit commands.
- Extends the merge trust gate to run the exact conditional Ix.Tc manifest in
  addition to the completed and statement frontiers.
- Narrows the exact upstream axiom manifest after the current Lean4Ix revision
  removed the transitive dependency on `Lean4Lean.ptrEqExpr_eq` from the
  produced-generation transaction roots.

### 2. Audited Ix.Tc verification checkpoint

The `Ix.Tc.Verify` tree connects the production checker to
`Lean4Lean.Theory` through explicit translation, state, cache, collision, and
world-extension invariants. The public checkpoint includes:

- successful standalone constant checking through the production
  `TcM.checkConst` path, including concrete promotion from pending to trusted
  semantic worlds;
- exact failure and rollback behavior for rejected checks, rather than an
  invariant that constrains only successful executions;
- coordinated block classification, all-or-nothing admission, ownership,
  block-cache publication, and the anonymous-environment driver;
- proof-visible cache, context, inference, WHNF, definitional-equality,
  projection, and bounded-recursion interfaces used by those public roots;
- a concrete supported Boolean environment whose production enumeration and
  checking schedule is related to an independently justified semantic
  topological schedule;
- singleton, indexed, mutual, recursive-Pi, alias, and nested inductive
  fixtures, including constructor positivity and validation traversals;
- generated-recursor construction, selection, metadata, type closure,
  member checking, rule comparison, and atomic family/recursor admission;
- the concrete `LeanBox`/`LeanTree` nested transaction, with restored
  recursors and rules, both production block checkers, exact restored iota
  patterns, and no surviving flat auxiliary declaration;
- small-elimination and `Eq`/K-target fixtures that preserve recursor universe
  layout and the generated K bit;
- an Ix projection relation instantiated by Lean4Ix's registered projection
  semantics, including substitution, universe instantiation, and a concrete
  dependent-record projection; and
- conditional quotient admission and quotient-reduction bridges that invert
  the real production executions while taking the still-upstream semantic
  transactions as explicit theorem inputs.

The production checker changes associated with this proof work also fail
closed when a coordinated block body disappears and avoid falling through to
a second standalone check after coordinated handling.

The inductive implementation is factored into proof-visible stages and kept
aligned with Rust. The main hardening points include checked metadata sums,
nested-application arity checks, specialization-aware auxiliary deduplication,
block-level recursor construction, complete-type selection for auxiliary
recursors, canonical peer-header checks, and exact type/rule/K-target
comparison before accepting stored recursors.

Complete-type selection first checks the already-validated canonical block
position and short-circuits only after a successful closed-type `DefEq` call.
If that candidate is absent, metadata-incompatible, or unequal, both Lean and
Rust retain an auditable finite fallback over every remaining candidate. The
positional branch and fallback callback domain are covered by the selection
state-preservation theorem. This avoids pathological comparisons between
wrong nested auxiliaries that share very large binder prefixes without
returning to the unsound open-major-domain heuristic.

This is not an unconditional claim about every Ix.Tc feature. The exact
completed manifest contains 2,034 theorem roots. A separate one-root
conditional manifest names the two remaining mutual-recursion witnesses:

- `Ix.Tc.Upstream.Pending.mutualTreePhysicalGenerationWF`
- `Ix.Tc.Upstream.Pending.mutualTreePhysicalRulePatternSound`

Quotient bridges similarly receive their semantic transaction/law resources
as parameters; Ix does not manufacture those resources with a new global
axiom. Remaining Lean4Ix metatheory debt is recorded by exact transitive
origin rather than duplicated or hidden locally.

### 3. Ixon v2 binder modes and wire format

Ixon expressions now carry substructural intent directly:

- lambdas store `Uses` (`erased`, `linear`, `affine`, or `many`);
- foralls store both `Uses` and `Owned` (`unique` or `shared`);
- ordinary Lean lambdas compile as `many`;
- ordinary Lean foralls compile as `many/shared`.

The stable format id is `ixon-v2`; `.ixe` environment version 2 begins with
the `0xE2` header. The Lean and Rust readers/writers agree on the new mode
bytes and maximal lambda/forall telescope grammar. They reject invalid mode
bits, noncanonical nested telescopes, invalid Boolean let flags, trailing
bytes in full-buffer decoders, and incompatible environment versions.

All affected in-repo consumers are ported, including production compilation,
Lean and Rust decompilation, anonymous and metadata ingress, sharing,
typechecking, FFI conversion, tests, and documentation. Existing v1 `.ixe`
artifacts must be regenerated; the version error reports that requirement
directly.

The Aiur IxVM is now a first-class v2 consumer as well. Its internal Ixon
schema carries the same `Uses` and `Owned` fields, its deserializer validates
the exact v2 mode ranges and Boolean let flag, and its reader rejects
non-maximal App, lambda, and forall telescopes. Its serializer emits the modes
again, so `Reveal` checks the address of the exact committed v2 expression;
the modes are erased only afterward when converting into the current
mode-agnostic kernel expression representation.

Because binder bytes participate in content addressing, the v2 transition
also rekeys canonical declarations. This checkpoint regenerates the canonical
Rust primitive-address table, the Lean typechecker table, every Aiur IxVM
primitive literal, the generated 5.8 MB Rust Aiur kernel, and the affected FFT
cost pins. The legacy/original primitive table and the synthetic
`eagerReduce` marker remain unchanged. Address-parity tests scan the elaborated
Aiur AST, including the formerly inline `Eq` address, so an unlabelled stale
literal cannot bypass the table comparison.

The IxVM coverage includes a positive hash-bound `Reveal` fixture containing
linear, affine, and erased usage plus unique/shared ownership. Raw committed
fixtures exercise the negative boundary for reserved mode bits, non-Boolean
let flags, and non-maximal App/lambda/forall telescopes.

The semantic bridge proves that erasing v2 modes is constructive in both
directions for the related ordinary Lean fragment. This PR does not yet make
the Lean compiler infer or exploit linear/affine/unique modes; its current
output deliberately inhabits the conservative `many/shared` subset.

### 4. Independent compiler semantics

This PR adds `IxCompileVerify` as a separate, non-default Lean library. Its
statement frontier deliberately does not import checker acceptance as the
compiler specification.

The new source/target model includes:

- `IxonExprRel`, a table-aware relation from Ixon expressions to Lean4Ix
  `VExpr` values;
- direct coverage for external references, mutual references, sharing,
  literals, lambdas, foralls, lets, metadata erasure, and projections;
- a `SourceExprRel` for named Ix source syntax and an explicit
  `KernelSourceWitness` boundary, without treating the production compiler or
  decompiler as a semantic oracle;
- a finite `Catalog` with explicit canonical, anonymous-operational,
  metadata-sidecar, and decompile-only views;
- `Catalog.WF`, table resolution, ordered sharing resolution, projection
  address integrity, and explicit digest/key-faithfulness assumptions; and
- a total ordinary-fragment reference compiler whose universe and expression
  outputs preserve the independent Lean4Ix value.

Hash-map and content-address assumptions are scoped to finite run support.
They are theorem premises, not implicit claims that cryptographic collision
freedom follows from Lean's `BEq` implementation.

### 5. Production compiler refinement

The production compiler now exposes kernel-visible total paths for the part
covered by this checkpoint:

- `CompileM.compileUniv` is structurally total and refines the reference
  universe compiler through `CompileM.run`.
- Universe interning preserves finite-table well-formedness, sound warm
  memos, and the independent Lean4Ix universe value.
- `CompileEnv.surgeryFree` selects a total `compileExprNoSurgery` path when no
  ordinary, `brecOn`, or `below` call-site plans are present.
- Plan-bearing environments retain the existing eta/plan-aware surgical
  compiler; the merge preserves main's plan-head arity checks and eta adapter
  behavior.
- The total expression path preserves the existing production flattened-App
  spine, arena allocation order, cache lookup/insertion behavior, local mutual
  references, external references, arbitrary universe vectors, literals,
  projections, binders, lets, and metadata. Lean/Rust equality remains checked
  by the differential gates below.

The refinement proof is layered rather than monolithic:

- structural expressions first refine the reference compiler with a sound
  warm expression cache;
- frozen preseed relations then cover table-backed external and mutual
  constants, universes, literals, and projections across the complete
  recursive ordinary-expression tree;
- `ArenaRel` connects the returned metadata root to the source expression;
- arena growth is append-only and bounded by a cache-free source allocation
  cost, with `UInt64` no-wrap stated explicitly; and
- all of the above composes into a production-run theorem preserving the
  independent Lean4Ix expression value.

The declaration pipeline is then connected to the codec domain through the
actual production state machine:

- a total structural preseed collector finds every reference and universe
  leaf needed by one or many roots, while an explicit active-ancestor
  collision premise accounts for digest/context seen-set hits;
- canonical reference and universe commits are proved executable, preserve
  their lookup maps, establish frozen-index completeness, and discharge
  table-capacity bounds from structural source costs;
- recursive sharing analysis, representative selection, usage propagation,
  rewriting, and sharing-table construction preserve expression wire safety;
- the common finalizer derives the root array from each compiled
  `ConstantInfo`, retains the final block state, and produces bytes that
  decode back to the exact block; and
- default-state production drivers are closed for axioms, definitions,
  theorems, opaque declarations, quotients, standalone recursors, inductives,
  constructors, and all four projection forms.

The true mutual path is covered as well. Heterogeneous preseeding resets the
universe memo for each member context while preserving the shared primary
tables. Definition, inductive, and recursor members satisfy one common frozen
state contract; alpha-equivalent classes retain one payload representative
but compile metadata for every member. The proof follows both the singleton
collapse and general mutual wrapper branches through projections, sharing,
serialization, and `BlockResult.mk'`.

Finally, the non-singleton `compileConstant` entry is decomposed into SCC
member lookup, canonical classification, and mutual compilation. The custom
monadic merge sort is totalized with explicit fuel; comparison readiness
constructs every expression and constant comparison; grouping preserves the
member count and produces nonempty classes; and each unfinished refinement
round strictly increases the number of classes. The source-member bound
therefore rules out fuel exhaustion. The exported theorem constructs both
collection and sorting runs from source evidence and no longer accepts raw
execution hypotheses for either phase.

Call-site surgery itself is not claimed by these theorems. The public
dispatcher is proved only under `CompileEnv.surgeryFree`; plan-bearing
behavior remains protected by the existing differential tests while its
formal refinement is future work.

### 6. Metadata and presentation-store refinement

Expression metadata is no longer restricted to empty maps in the verified
ordinary path.

- Production `serializeIxSyntax` is structurally total and has an exact
  reference/run theorem.
- Metadata compilation covers strings, booleans, names, naturals, integers,
  source spans, pre-resolved identifiers, nested syntax arrays, and final
  serialized syntax blobs.
- `compileDataValue` and `compileKVMap` refine their total reference encoders
  while changing only presentation-side name/blob stores.
- A separate strict theorem scopes name-hash and blob-address faithfulness to
  the finite metadata traversal, preserves old lookups, and proves recovery
  of every collected name, ancestor name component, and blob payload.

This separation is intentional: anonymous semantic preservation does not
inherit stronger metadata-key assumptions merely because a source expression
contains presentation data.

### 7. Complete Ixon expression and constant codecs

The production Lean universe and expression writers/readers are
kernel-visible total definitions, and the exact inverse proof now spans the
full explicit compiler-facing wire domain.

For universes, this PR proves:

- exact full-buffer decoding;
- inverse laws for trimmed little-endian integers and both `Tag2` forms; and
- exact round trips for the full representable universe domain, including the
  `Sort 1` universe required by the first declaration fixture.

For expressions, this PR proves:

- exact `Tag0` and `Tag4` inverse laws;
- full-buffer rejection of trailing bytes; and
- exact round trips for all twelve expression constructors, arbitrary
  wire-sized reference universe vectors, and arbitrary canonical maximal
  App/lambda/forall spines.

The constant layer composes those laws through every production payload:

- axiom, definition, theorem, opaque, quotient, and standalone recursor
  declarations;
- constructors, inductives, all four projections, and definition projections;
- recursor rule arrays, packed flags, constructor arrays, and all three
  `MutConst` member tags; and
- arbitrary wire-representable sharing, reference, and universe tables.

`Constant.wireWF` is proved equivalent to the codec's compositional domain,
including lossless count conversion, exact 32-byte addresses, and
representable compressed universe successors. Consequently every explicit
wire-safe `ConstantInfo` value—including counted mutual blocks—round-trips
through the production `serConstant`/`deConstant` pair with exact
full-buffer consumption.

### 8. Exact trust manifests

Trust is checked per exported theorem root rather than summarized informally.

- `Ix.Tc.Verify.Audit.Completed`: 2,034 completed roots.
- `Ix.Tc.Verify.Audit.Conditional`: one root with exactly named pending
  dependencies.
- `Ix.Tc.Verify.Audit.Statements`: seven statement/frontier roots.
- `Ix.Compile.Verify.Audit.Statements`: 143 compiler roots.
- Both Ix.Tc and compiler source-frontier audits reject any local declaration
  that directly uses `sorryAx`.

Each root records its exact standard axioms, native axioms, upstream
`sorryAx` origins, pending assumptions, and forbidden dependencies. The audit
fails both when an unlisted dependency appears and when a stale expected
dependency disappears. The compiler statement frontier independently relates
source and Ixon values and does not use checker acceptance as its semantic
specification.

The conditional Ix.Tc audit is added to the merge workflow. `IxCompileVerify`
remains deliberately separate and non-default, so its focused build/audit
command is included below.

## Compatibility and operational impact

- **Wire compatibility:** `.ixe` v1 artifacts are not accepted as v2 and must
  be recompiled.
- **Content addresses:** any expression or declaration whose canonical bytes
  contain binders receives a new v2 address. Canonical primitive tables and
  fixtures in this repository have been regenerated; external catalogs,
  claims, caches, and address-pinned artifacts must be rebuilt.
- **Compiler output:** ordinary Lean compilation remains conservative
  (`many/shared`) and passed Lean/Rust byte-differential checks.
- **Call-site plans:** plan-bearing compilation continues through the existing
  surgical implementation; main's eta adapters and plan-head validation are
  preserved.
- **Kernel behavior:** malformed nested applications, inconsistent generated
  recursor peers, missing generated recursors, and mismatched recursor
  metadata/rules now fail closed. Generated recursor selection retains full
  closed-type validation while checking the canonical block position before
  the finite fallback scan.
- **IxVM behavior:** Aiur parses and reserializes the exact v2 expression at
  the content-address boundary, rejects noncanonical encodings, and erases
  binder modes only when entering the current kernel AST.
- **Cryptography and proof artifacts:** BLAKE3, SHA-256, Merkle, claim, and
  multi-STARK algorithms and codecs are unchanged. Their inputs are not:
  regenerated Ixon addresses and the changed guest/kernel programs invalidate
  cached Aiur proofs and any SP1/Zisk proof or verification-key material tied
  to the prior program bytes. Those artifacts must be regenerated rather than
  translated.
- **Build topology:** the large proof libraries remain outside the default
  runtime build. They are built explicitly for formalization and trust
  auditing.
- **Trust posture:** no claim silently upgrades an explicit collision,
  native-reduction, upstream theorem, or semantic-transaction premise into a
  proved fact.

## Validation

The branch was merged with current `main` before final validation. The merged
compiler checkpoint passed the default build, exact compiler trust audit, and
complete flake evaluation:

### Post-merge recursor-selection regression

```sh
nix develop --command lake build \
  Ix.Tc.Verify.Inductive.GeneratedRecursorSelection
nix develop --command lake build \
  Ix.Tc.Verify.Audit.Completed \
  Ix.Tc.Verify.Audit.Conditional \
  Ix.Tc.Verify.Audit.Statements \
  Ix.Tc.Verify.Audit.SorryFrontier
nix develop --command cargo test -p ix-kernel
nix develop --command lake test --wfail -- --ignored \
  shard-map catalog-dedup commit-io fidelity-initstd \
  kernel-ixon-roundtrip kernel-tutorial kernel-check-env \
  kernel-check-const rust-kernel-build-primitives \
  rust-kernel-build-prim-origs ixvm
```

Result: the focused selection proof built with all 289 dependencies. The
post-merge trust audit passed all 2,034 completed roots, one conditional root,
and seven statement roots, and confirmed that the verification frontier has
no source declaration using `sorryAx`. Rustfmt passed; all 682 `ix-kernel`
unit tests passed; and the complete requested ignored suite passed. The
focused kernel check accepted all 79 constants, including
`Lean.Meta.Grind.Arith.Cutsat.CooperSplitPred.rec` and `Lean.Doc.Block.rec` in
68 ms and 8 ms respectively. The full environment accepted all 220,995
constants, the Ixon round trip checked 170,153 constants with zero errors, and
all IxVM execution, FFT-cost, exploit, and native-parity gates passed.

### Final compiler checkpoint

```sh
nix develop --command lake build
nix develop --command lake build Ix.Compile.Verify.Audit.Statements
nix flake check
git diff --check
```

Result: all commands passed. The default build completed 301 jobs; the trust
audit checked 143 exact compiler roots; the compiler-verification frontier
contains no `sorry` or `admit`; and the flake check reported only the expected
notice that checks for incompatible systems were omitted.

The underlying Ixon v2/IxVM integration checkpoint, preserved as
`131894c2`, also received the following subsystem validation before the final
compiler-proof commit.

### Default build

```sh
nix develop --command lake build
```

Result: passed, 301 jobs.

### Ix.Tc proof and trust frontier

```sh
nix develop --command lake build \
  Ix.Tc.Verify.Audit.Completed \
  Ix.Tc.Verify.Audit.Conditional \
  Ix.Tc.Verify.Audit.Statements \
  Ix.Tc.Verify.Audit.SorryFrontier
```

Result: passed, 620 jobs; 2,034 completed roots, one conditional root, seven
statement roots, and no Ix.Tc verification source declaration using
`sorryAx`.

### Compiler proof and trust frontier

```sh
nix develop --command lake build Ix.Compile.Verify.Audit.SorryFrontier
```

Result: passed; no compiler-verification source declaration uses `sorryAx`.

### Lean/Rust auxiliary compiler differential

```sh
nix develop --command lake test -- --ignored aux-gen-diff
```

Result: passed all plain-drift, expansion, patch, plan, driver, and parallel
gates. Lean and Rust produced identical 6,893,510-byte environments.

### Ingress corpus

```sh
nix develop --command lake test -- --ignored tc-ingress-meta
```

Result: passed all seven corpus groups: `nat-add`, `list-map`, `nat-arith`,
`eq-basics`, `bool-decide`, `char-ordinal`, and `inductives-recursors`.

### Decompiler differential

```sh
nix develop --command lake test -- --ignored decompile-diff
```

Result: passed for 6,615 constants with zero errors or mismatches: 3,969 plain
constants and 2,646 auxiliary constants. Call-site replay and plan parity
gates also passed.

### Ixon codec and parity suite

```sh
nix develop --command lake test -- ixon
```

Result: passed the full Ixon syntax/codec suite, including v2 mode vectors,
malformed-input rejection, and Lean/Rust byte parity.

### IxVM generated kernel and security boundary

```sh
nix develop --command lake exe ix codegen --check
nix develop --command lake test -- prim-addrs primitive-address-parity
nix develop --command lake test -- --ignored ixvm
```

Result: generated Aiur sources were current; canonical and original primitive
address parity passed across Rust, Lean, and 89 named Aiur literals; and the
full ignored IxVM suite passed. The latter includes all kernel execution and
interpreter/native parity cases, every regenerated FFT cost pin, the shard
pipeline pin (`6,859,583,032`), the positive v2 binder-mode `Reveal`, and all
six malformed-v2 rejection fixtures.

### Cryptography and recursive-verifier regression suite

```sh
nix develop --command lake test -- \
  aiur-hashes claim merkle assumption-tree multi-stark
```

Result: passed BLAKE3 and SHA-256 execution/proving vectors, claim and
assumption-tree codecs, Merkle properties, and the multi-STARK PCS, challenger,
FRI, reduced-opening, and field-arithmetic reference comparisons.

### Rust formatting, lint, unit tests, and workspace check

```sh
nix develop --command cargo clippy --workspace --all-targets --all-features
nix develop --command cargo check --all-targets --all-features --workspace
nix develop --command cargo test -p ixon
```

Result: rustfmt and clippy passed cleanly; all 324 active `ixon` unit tests
passed (three diagnostic dump tests ignored); and the complete host workspace
compiled, including `ixon`, `ix-kernel`, `aiur`, `multi-stark`,
`ixvm-codegen`, `ix-compile`, and the FFI crates.

The SP1 and Zisk guests depend on the shared `ixon`, `ix-kernel`, and
`ix-common` crates, so their CI jobs must rebuild their guest ELFs and derive
fresh program verification keys. Those specialized guest toolchains were not
run locally; the ordinary Rust workspace check does not compile the excluded
zkVM targets.

The final head contains no unresolved Jj conflict markers.

## Suggested review order

The branch is intentionally split into reviewable commits:

| Commit | Review focus |
| --- | --- |
| `8a17a720` | Ix.Tc verification checkpoint, production checking/inductive changes, and trust manifests |
| `49f2dd43` | Ixon v2 binder modes and Lean/Rust representation port |
| `f5660b1b` | Conservative v2 value relation to Lean4Ix |
| `d1eae96b` | Reference compiler semantics and production state relation |
| `6fece592` | Production expression compiler refinement |
| `5a077b94` | Table-backed leaves and universe/reference interning |
| `5cc3d4af` | Complete ordinary-expression refinement and value theorem |
| `cac72b91` | Metadata arena relation and growth bounds |
| `2f205d5f` | Scalar expression metadata integration |
| `fa449102` | Total syntax metadata serialization |
| `fd1a7417` | Strict metadata side-store recovery |
| `fee3b893` | Initial total universe codec laws |
| `df7d67cb` | Full representable universe codec theorem |
| `aa286acc` | Singleton-domain expression codec theorem |
| `6f18ea90` | Merge of current main and compiler/ingress reconciliation |
| `131894c2` | Aiur IxVM Ixon v2 consumer, regenerated addresses/kernel, strict Reveal tests, FFT pins, and Rust lint cleanup |
| `066b32bc` | Complete expression/constant codec domain and codec-safe production singleton, inductive, recursor, sharing, and mutual compilation |
| `72e1140c` | Clean merge of current `main`, including independent TruthMines-library validation and opt-in CUDA proving |
| `680ebbf8` | Verified canonical-position recursor fast path, exhaustive fallback, diagnostics, and full-environment regressions |

A practical review sequence is:

1. review Ixon v2 and production consumer changes;
2. review the public Ix.Tc and compiler statement frontiers;
3. inspect the exact trust manifests and forbidden-dependency lists;
4. review proof modules by theorem slice; and
5. use the differential and corpus gates to assess production equivalence.

## Non-goals and follow-up work

This PR does not claim:

- complete Ix.Tc soundness for every kernel feature and every hostile Ixon
  environment;
- constructive discharge of the separately named mutual-recursion or
  quotient semantic resources still assigned to Lean4Ix;
- removal of all upstream Lean4Ix metatheory debt;
- a production compiler for non-conservative linear, affine, erased, or unique
  binder modes;
- formal refinement of plan-bearing call-site surgery;
- source-to-target semantic preservation for declarations and mutual blocks
  (X3), beyond the codec-safety and exact-round-trip result proved here;
- whole-environment semantic preservation or canonical-address invariance
  (X4-X5);
- full metadata/decompile/recompile recovery (X6);
- refinement of the parallel Lean compiler or Rust compiler to the verified
  reference pipeline (X7); or
- Rust-kernel, Aiur, or proof-system soundness.

The next bounded compiler slice is to add declaration denotations to the
existing source/Ixon value relation and lift the now-verified production
singleton and mutual drivers from codec safety to source-preserving X3
theorems. That semantic layer can then feed whole-environment preservation and
canonical-address invariance.
The typechecker track can continue independently by replacing the remaining
conditional Lean4Ix resources with constructive definitions and proofs as
they become available.


arthurpaulino:
!benchmark aiur-sharded-env

samuelburnham (dismissed):

arthurpaulino:
!benchmark aiur-sharded-env

arthurpaulino (approved):

PR #597, opened 2026-08-27
Add opt-in CUDA acceleration for Aiur proofs

arthurpaulino:
Expose multi-stark's first-party CUDA prover through Aiur and the Ix FFI while keeping ordinary Cargo, Lake, and Nix builds CPU-only and independent of the CUDA toolkit. Enable it explicitly with IX_CUDA=1, IX_CUDA=true, or IX_CUDA=yes for Lake commands, or with the cuda Cargo feature for Rust consumers.

Forward the feature through aiur and ix-ffi and keep Cargo.lock synchronized with the selected multi-stark revision. Merge the multi-stark PR first, then replace the temporary topic-branch pin with the resulting main-branch revision before merging Ix; Ix must not ship with a long-lived topic-branch dependency.

Make Lake feature selection part of the Rust archive dependency trace and snapshot feature-specific artifacts. Switching IX_CUDA now rebuilds and relinks lake exe bench-typecheck directly instead of silently retaining the backend from a previous build, and the selected Rust feature set is printed during the build.

Contain prover panics, including CUDA initialization and runtime failures, at every Lean proving FFI boundary and return ordinary Except String errors instead of unwinding across the C ABI and aborting the Lean process. Update CLI, benchmarks, and tests for the explicit error path.

Restore parallel, net, and test-ffi coverage in ordinary release clippy/check CI. Keep CUDA out of Nix's host-feature aggregation, include the Rust trace inputs required by Lake's prebuilt archive target, and add a toolkit-only CUDA job that compiles the Rust workspace and links a CUDA-enabled Lean benchmark executable without requiring a GPU. Move hardware-specific performance numbers out of the stable README and document the reproducible Vector.extract_append recursive workflow.

On an RTX PRO 6000 Blackwell, Vector.extract_append recursive q50 reduces the post-execution-and-trace portion of inner and outer proving from 71.87s on CPU to 8.74s with CUDA (8.22x), with unchanged proof sizes and successful verification. End-to-end proving also includes Aiur execution and trace construction, which remain outside this comparison. Fifty queries is a development benchmark parameter, not a production security recommendation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,9 +101,9 @@ jobs:
       - name: Check Rustfmt code style
         uses: actions-rust-lang/rustfmt@v1
       - name: Check clippy warnings
-        run: cargo clippy --workspace --all-targets --all-features
+        run: cargo clippy --release --workspace --all-targets --features ix-ffi/parallel,ix-ffi/net,ix-ffi/test-ffi -- -D warnings
       - name: Check *everything* compiles
-        run: cargo check --all-targets --all-features --workspace
+        run: cargo check --release --workspace --all-targets --features ix-ffi/parallel,ix-ffi/net,ix-ffi/test-ffi
       - name: Tests
         run: cargo nextest run --release --profile ci --workspace
       - name: Get Rust version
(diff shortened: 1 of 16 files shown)


arthurpaulino:
!benchmark

arthurpaulino:
!benchmark

arthurpaulino:
!benchmark

arthurpaulino:
!benchmark fresh

johnchandlerburnham (dismissed):

johnchandlerburnham (approved):

PR #612, opened 2026-09-04
Update multi-stark for hybrid CUDA scheduling

arthurpaulino:
Point Aiur at the multi-stark revision that balances large mixed-height commitments across host CPUs and CUDA according to device capacity. The backend carries hybrid placement through lookup, quotient, openings, and FRI while preserving the existing Goldilocks/BLAKE3 protocol and CPU verification path.

No Ix source or build configuration changes are required: the existing opt-in cuda feature chain selects the new backend, while default builds remain CUDA-independent. Cargo.lock records the matching dependency graph.

Validated with release clippy across the non-CUDA CI feature set, a release all-targets CUDA check, and IX_CUDA=1 lake build bench-typecheck, including the final Rust static-library link.


arthurpaulino:
!benchmark fresh

johnchandlerburnham (dismissed):

PR #619, opened 2026-09-08
Function groups

gabriel-barrett:
# Aiur function grouping: prove several functions with one circuit (IxVM proofs −50–56%)

An IxVM proof commits and opens only the circuits active in the run, and every
column of an active circuit is opened at every FRI query. A kernel check
activates 500–575 of the kernel's 744 function circuits, most of them a few
rows tall, for a summed active width of 21–24k columns — so proof size is
dominated by *how many* circuits are active, not by the hot ones. This PR lets
several functions share one circuit, and uses that to cut IxVM proofs roughly
in half at +1–2% prover cost.

## Mechanism (`aiur: circuit-level function grouping`)

A circuit can back several functions. Members are walked like the branches of
a single function: auxiliary columns and lookup slots are shared across members
(the same save/restore sharing match arms already use), selector columns are
laid out consecutively per member, and every member folds its selector-gated
return message (with its own function index) into the shared lookup slot 0
against one multiplicity column. One extra constraint enforces cross-member
exclusivity: the sum of the members' top-block selectors must be boolean.

Grouping is a circuit-level choice, not a property of the function library:
callers still target function indices on the function channel, so execution,
the query record, the interpreter and codegen are untouched.
`Source.Toplevel.compile` builds the default singleton partition
(`Bytecode.Toplevel.circuits`, behavior-identical to before) and
`CompiledToplevel.groupFunctions` regroups it by function name (validated:
known, constrained, non-entry, no duplicates). The merged layout is max inputs,
summed selectors, max auxiliaries, max lookups. Rust consumes the partition
directly via FFI; constraints, trace and synthesis iterate circuits, and
witness rows concatenate the members' queried rows in member order.

## Wiring (`aiur: wire optional function-grouping application points`)

Every site that compiles a production toplevel for proving or verifying goes
through `compileWithGroups` with a per-toplevel datum:

- `IxVM.functionGroups` — the kernel (check/prove/verify/refine/aggregate, the
  batch checker, the ixvm test runner, the benches)
- `Aggr.functionGroups` — the `ix_aggr` recursion system (**empty**)
- `MultiStark.verifierFunctionGroups` — the standalone verifier (**empty**)

Members are resolved by the exact string the statistics print, so measured
groupings feed back verbatim. The recursion groupings stay empty on purpose.

## The IxVM grouping (`ixvm: populate the function grouping …`)

`function-groups/ixvm/` holds the profiles, the model, the partitioner and a
README with the method. In short:

- **Model.** Committed width and the FFT cost model of
  `Ix/Aiur/Statistics.lean` are reproduced from the static layout alone,
  validated to the column on every circuit, so a candidate grouping's widths
  and costs are known before building anything.
- **Partitioner.** Greedy agglomerative merging. Benefit = active committed
  width saved on the training workloads (String.split, Array.extract_append,
  and a 2^16384 big-nat reduction whose hot circuit is cold everywhere else).
  Cost = modelled prover cost added (FFT model plus a constraint-evaluation
  term), relative to each workload. Merges are taken by benefit per unit cost
  within an average 2% (max 4%) increase per workload, ≤ 16 members and ≤ 40
  selectors per group.
- **Result.** 83 groups over 656 circuits: 744 → 171 circuits, total
  committed width 33,693 → 12,729.

Measured on the grouped kernel (held-out constants included):

| workload | active circuits | active fn width | FFT cost |
|---|---|---|---|
| String.split (train) | 575 → 171 | 24,094 → 10,163 (−57.8%) | +1.0% |
| Array.extract_append (train) | 572 → 171 | 23,884 → 10,163 (−57.4%) | +1.1% |
| nat_pow_big (train) | 461 → 163 | 19,688 → 9,452 (−52.0%) | +1.9% |
| Nat.sub_le_of_le_add | 503 → 167 | 21,332 → 9,870 (−53.7%) | +1.6% |
| String.Internal.append | 494 → 166 | 20,937 → 9,941 (−52.5%) | +1.9% |
| Lean.Syntax.rec | 509 → 174 | 21,886 → 10,942 (−50.0%) | +2.0% |

End to end (`bench-typecheck --ixe InitStd.ixe`, 100 queries):

| constant | proof bytes | prove | verify |
|---|---|---|---|
| Nat.sub_le_of_le_add | 19,765,334 → 9,414,761 (−52.4%) | 2.79s → 2.32s | 0.103s → 0.051s |
| Lean.Syntax.rec | 20,233,436 → 10,325,344 (−49.0%) | 3.06s → 2.47s | 0.099s → 0.054s |
| String.split | 22,452,866 → 9,895,153 (−55.9%) | 21.49s → 20.32s | 0.121s → 0.056s |

Proving got slightly faster despite the modelled +1–2%: the per-circuit fixed
costs of the ~570 shed circuits outweigh the extra columns on the few-row
merged traces. For comparison, the earlier shape-band heuristic on the old
`group-functions` branch bought −43% active width for +9–14% FFT cost.

## Tests

- `aiur-prove` proves the same toplevel ungrouped and with a 3-member test
  group (different arities, matches, cross-member call, recursion), plus
  structural checks on the partition.
- Kernel FFT pins re-measured (median +1.7%, max +3.8%; shard pipeline
  +1.3%). The `ixvm` ignored suite, the primary suite (including
  `recursive-verifier`, `ix-aggr`, `aggregate-first`), the Rust `aiur` tests,
  fmt, clippy and `ix codegen --check` are green.
- The Mathlib Stage 2 fixture stays fenced as obsolete (it already predates the
  protocol bump; the IxVM verifying key changes again here).

## Notes

- The IxVM verifying key changes; existing IxVM proofs will not verify against
  this build.
- A stale grouping stays sound (grouping never affects semantics), only less
  efficient, and fails loudly if a member name disappears. Regeneration steps
  are in `function-groups/ixvm/README.md`.


arthurpaulino:
!benchmark fresh

arthurpaulino (approved):

PR #620, opened 2026-09-09
recursion: populate the ix_aggr and verifier function groupings (192 …

gabriel-barrett:
# Recursion function groupings: ix_aggr 192 → 39 circuits, verifier 191 → 72

Stacked on #<function-groups PR>. Applies the same cost-aware grouping to
the two recursion toplevels that the base PR left empty.

Groups were built from profiles of the `ix_aggr` join of two lifted shards
and of the standalone verifier over two IxVM proofs (Nat.sub_le_of_le_add,
Lean.Syntax.rec), ranking merges by active committed width saved per unit
of modelled prover cost relative to each workload, within an average 2%
(max 4%) increase, ≤ 16 members per group. The standalone verifier gets
the same groups restricted to its own functions.

Measured with `bench-typecheck --recursive --join`:

| system | circuits | FFT cost | proof | prove | verify |
|---|---|---|---|---|---|
| verifier over Nat.sub_le_of_le_add proof | 191 → 72 | +1.5% | 4.19 → 2.31 MB | 30.5s → 30.3s | 24 → 14 ms |
| verifier over Lean.Syntax.rec proof | 191 → 72 | +1.4% | 4.20 → 2.31 MB | 30.7s → 30.6s | 25 → 14 ms |
| ix_aggr join of the two | 192 → 39 | +1.4% | 4.98 → 2.42 MB | 58.8s → 58.4s | 28 → 14 ms |

Active width drops ~55% on all three; the FFT increase stays under budget
because the verifier's cost sits in a few tall circuits (blake3, FRI
folding) the partitioner leaves alone. The `ix_aggr` verifying key
changes; existing aggregate proofs will not verify against this build.

Tests: primary suite (`recursive-verifier`, `ix-aggr`, `aggregate-first`,
`multi-stark`), the `aggregate-activation` audit, and `ix codegen --check`
pass. No codegen change.

### DIFF
--- a/Ix/Aggr/FunctionGroups.lean
+++ b/Ix/Aggr/FunctionGroups.lean
@@ -13,7 +13,227 @@ public section
 
 namespace Aggr
 
-def functionGroups : Array (String × Array String) := #[]
+-- 21 groups over 174 circuits, cost-aware greedy merging
+-- over the ix_aggr join and two standalone-verifier profiles (average 2%,
+-- max 4% modelled cost increase per workload).
+def functionGroups : Array (String × Array String) := #[
+  ("aggr_group_00", #[
+    "memo_u32_less_than",
+    "u64_is_zero",
+    "relaxed_u64_succ",
+    "flatten_u64",
+    "read_sys_lookups_n",
+    "frontier_merge",
+    "frontier_split",
+    "frontier_sort",
+    "frontier_leaves",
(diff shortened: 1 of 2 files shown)


gabriel-barrett:
!benchmark fresh

johnchandlerburnham (approved):

PR #624, opened 2026-09-10
perf(ixvm): avoid repeated substitution and row copying in FLT checks

arthurpaulino:
Two targeted changes address the large NumberField, Weierstrass and RubinSilverberg proofs from AnthropicFLT:

- Keep checked let types and delayed values in a local inference environment, materializing terms only when needed. Preserve context-sensitive memo keys; whnf and definitional equality still operate on ordinary expressions.
- Build multiplication rows with Cons instead of repeatedly copying a growing output prefix. Preserve limb arithmetic, carry checks and exact results.

Remove the obsolete function group referencing the eliminated limb-list snoc helper, regenerate the native executor, and refresh the existing IxVM FFT pins. No diagnostic stubs or ad hoc test runners are included.

Validation:
- All three focused FLT subjects finish within the 600s bytecode cap on the pre-grouping base (218s / 553s / 472s). These were subject-only diagnostics with hash/dependency checks temporarily disabled, not full FLT validation.
- On main at 4c912543, all 64 ISLB shards pass with full verification: 183,062 owned constants, 32 workers, 89.78s wall time.
- Rebuilt IxTests; lake test -- --ignored ixvm passes in 3m27s after repinning, including native/bytecode parity. Codegen consistency checks pass.

This is a targeted performance tradeoff: 49 constant pins decrease and 30 increase (largest +2.11%); the shard-pipeline pin is refreshed too.


arthurpaulino:
!benchmark fresh

johnchandlerburnham (approved):

PR #627, opened 2026-09-11
perf(ixvm): reduce FLT checking costs and witness memory

arthurpaulino:
Recover the remaining FLT checks without reverting delayed-let inference or tuning reduction fuel.

Changes:

- Strengthen delayed application contexts across unused Let/Local prefixes. Count raw bindings separately from type-context slots, lift the inferred type back, and keep introductions and pure-let contexts delayed.
- Stop minimum loose-variable-index scans at zero, preserving the exact summaries and memo keys while skipping irrelevant subtrees.
- Compare projected function heads through definitionally equal scrutinees before unfolding their potentially large arguments.
- Multiply U64 limbs in radix 2^16. Range-check decomposition advice and constrain reconstruction below the field modulus; reuse the checked high limb or its modular successor when propagating a one-bit carry.
- Fill pre-sized witness arenas directly in parallel instead of retaining converted partial buffers and copying growing arenas. Preserve stable offsets, channel contents and hints; test against serial construction.

Add two small synthetic FFT guards, prototyped with direct ix check runs: context strengthening costs 514,164,080 versus 808,853,595 with that path disabled; varied multi-limb multiplication costs 266,740,297 versus 277,782,128 with the previous arithmetic. Retain the existing let/row guards, regenerate the native executor, and refresh all affected suite FFT pins.

Native validation with full hashing and closure checking, no proving:

- Init+Std+Lean+Batteries: 64/64 shards pass; 14,154,767,400,043 total FFT.
- Mathlib: 256/256 pass; 147,662,851,272,684 total FFT.
- Anthropic FLT: 4099/4099 pass; 1,371,135,736,576,741 total FFT.

FLT validation refines overloaded shard 504 into four children, placing its four expensive certificates separately. Exact/disjoint coverage, unchanged external assumptions, and all 4095 untouched claim digests were verified. Shards 2083, 2747 and 3949 passed isolated retries after timeouts or host memory pressure during the concurrent sweep. This is not a claim that all original shards finish under the original concurrent schedule.

Against main 5430a5d9, the same 4077 successful FLT claims fall from 1,539,563,003,720,477 to 1,331,530,648,023,083 FFT (-13.51%). All remain successful: 4036 are cheaper and 41 more expensive. Do not compare the full new total with the incomplete baseline subtotal as though coverage matched.

The rebuilt IxVM suite passes its exact FFT pins, adversarial rejection checks, and native/bytecode parity checks. Temporary diagnostic stubs and CheckCmd fixtures are removed; environment/shard files and sweep logs stay outside the commit.


arthurpaulino:
!benchmark fresh

johnchandlerburnham (approved):

PR #631, opened 2026-09-13
fix(aiur): reject forged results and cyclic call traces

johnchandlerburnham:
Aiur could accept supplied traces for false public results or recursive calls
with no finite execution. This PR repairs four demonstrated trace defects
and a native Merkle commitment-binding defect. Regression tests construct
the offending witnesses and openings directly.

The implementation also reduces the cost of the repaired system through
checked call-order specialization, narrower rank and byte lookups, BLAKE3
and arithmetic changes, and smaller host data structures. Across 83 kernel
fixtures, summed raw FFT cost is **41.17% below the initial repair and
1.72% below original main**. Proving time and memory vary by workload.
In its matched comparison, BLAKE3 unrolling improves `Vector.append` proving
time at the cost of **19.93–34.40% larger proofs** across five claims and
slower verification.

## Soundness repairs

1. **Inactive rows cannot supply results.** Enforce
   `multiplicity * (1 - selector) = 0` on function and memory rows. This
   rejects the supplied witness for `3 * 5 = 16` and covers every grouped
   member. It adds a quadratic constraint without widening traces or lookups.
   [Regressions](https://github.com/argumentcomputer/ix/blob/585cea90742f428e7ca77c2d52c8ab09d32befb0/crates/aiur/src/synthesis/tests/acceptance.rs).

2. **Calls require a well-founded order.** Every constrained call advances a
   checked static component order, increases a bounded 48-bit rank within a
   recursive component, or satisfies a checked unit-step relation on a
   self-edge. This rejects self-supporting and mutually recursive witnesses.
   Honest recursion, shared callees and advice promotion remain supported.
   [Regressions](https://github.com/argumentcomputer/ix/blob/585cea90742f428e7ca77c2d52c8ab09d32befb0/crates/aiur/src/synthesis/tests/call_order.rs).

3. **Lookup messages have checked shapes.** Validate public entry visibility,
   claim and call input/output arities, and continuation yields. This prevents
   zero padding from interpreting an empty return's rank seven as output
   seven. Entry shapes are cached at construction; these guards add no AIR
   columns or FFT work.
   [Regressions](https://github.com/argumentcomputer/ix/blob/585cea90742f428e7ca77c2d52c8ab09d32befb0/crates/aiur/src/synthesis/tests/lookup_shapes.rs).

4. **Only terminal single-function circuits omit lookup argument gates.**
   Having one selector was insufficient: an empty branch could still write a
   store lookup and change a live message's channel. Branching circuits now
[...]

PR #634, opened 2026-09-15
fix(aiur): correct lookup constraints and strengthen proof validation

arthurpaulino:
Aiur could accept false public results through inactive provider rows, incorrectly combined lookup messages, and ambiguous function-message boundaries. The lookup-gating bug also prevented an honest execution from verifying. This PR fixes those failures, rejects Merkle-cap configurations that fail to bind every trace matrix, and adds explicit structural validation at the Aiur boundary.

The changes distill the first and third commits of [#631](https://github.com/argumentcomputer/ix/pull/631) into seven atomic commits. Call-rank machinery, cyclic-trace defenses, and performance improvements from that PR are excluded. Cyclic call traces remain outside this PR's scope.

The fixes distinguish **soundness** (rejecting proofs of false claims) from **completeness** (allowing honest executions to prove and verify):

| Commit | Change | Classification and evidence |
| --- | --- | --- |
| `2897d6b5` | Require function and memory provider multiplicities to vanish on inactive rows, including grouped function circuits. | **Soundness.** A supplied-trace regression previously proved `3 * 5 = 16`. Additional constraint tests cover inactive memory rows and every grouped member. |
| `9d5d0fe9` | Correct lookup argument gating and grouping for circuits with empty branches. | **Soundness and completeness.** An inactive branch's store could redirect a live function call into memory and establish a false result; an honest execution also failed verification. Both regressions are reproduced and fixed. |
| `c852ccce` | Validate constrained-call arities, continuation yields, and public claim channel, visibility, and arity. | **Soundness.** Lookup zero-padding allowed an empty return to establish a public claim containing an extra zero output. The regression uses the existing claim encoding without rank fields. |
| `15be618e` | Validate control counts, circuit members, and selector capacity before synthesis. | **Construction validation.** The malformed fixture already failed downstream; this makes structural bounds explicit and rejects it earlier. No new soundness or completeness failure is demonstrated. |
| `e6046b4e` | Require fixed preprocessed circuits to be active at their exact heights. | **Soundness hardening.** Tests establish early rejection of omitted byte tables and incorrect heights. These malformed proofs already failed downstream. |
| `58a6ce66` | Bound total lookup consumers below the field characteristic, with checked arithmetic and metadata alignment. | **Soundness hardening.** Makes the lookup-count assumption explicit. Tests cover field and machine bounds; they do not demonstrate a previously accepted false claim. |
| `e9048e89` | Reject caps that omit shorter trace matrices. | **Soundness.** Native MMCS tests demonstrate identical commitments for different short matrices and acceptance of altered single and multi-openings. Aiur rejects the unsafe cap/trace combinations and accepts the safe boundary. |

The adversarial regressions use the existing Rust Aiur test infrastructure and supplied trace rows, independently of honest witness generation. The Merkle-cap guard is also checked against direct injection geometry across 288,225 boundary cases. Honest proofs using unsafe cap configurations are intentionally rejected; the normal root-cap setting remains supported.

The lookup-gating fix has a conservative proving-cost tradeoff: only a single-selector, single-function circuit with terminal control retains raw lookup arguments and grouped accumulators. Some safe functions containing matches therefore lose that optimization. Execution counts and main trace widths are unchanged, but stage-2 width and potentially quotient degree increase. Measured FFT estimates rise by **0.22%** for `Nat.add_comm` and **0.38%** for the shard pipeline. The corresponding exact pins are updated in the gating commit. These are FFT-model changes, not measured wall-clock slowdowns; a more precise lookup-writer analysis is deferred.

Validation passed:

- All **35 Rust Aiur tests**: `cargo test --locked --release -p aiur --features parallel`.
- Strict workspace Clippy with CI features: `cargo clippy --locked --release --workspace --all-targets --features ix-ffi/parallel,ix-ffi/net,ix-ffi/test-ffi -- -D warnings`.
- Formatting: `cargo fmt --all -- --check`.
- Codegen parity: `lake exe ix codegen --check`; all three generated programs are unchanged.
- Lean suites: `lake test --wfail -- aiur-cross aiur-cost aiur-prove aiur-hashes rbtree-map multi-stark recursive-verifier ix-aggr`.
- Full IxVM suite, including generated-code parity and updated FFT pins: `lake test --wfail -- --ignored ixvm`.


arthurpaulino:
!benchmark fresh

gabriel-barrett (approved):

PR #635, opened 2026-09-16
perf(aiur): reduce witness memory and proving FFT costs

arthurpaulino:
Reduce Aiur's witness memory and modeled proving FFT work by extracting the independent optimizations from #631, on top of the correctness fixes merged in #634.

This PR contains six Aiur optimizations and a measured FFT-pin refresh, in seven atomic commits:

- **Cache claim shapes.** Summarize function return arities during system construction and reuse them for call validation and public claim checks, preserving the existing acceptance semantics.
- **Tune lookup grouping.** Choose accumulator groups using compiled expression degrees and the commitment parameters. Require no increase in modeled FFT work at any integer trace height, including small traces, and keep the circuit metadata consistent with the chosen grouping.
- **Compact witness metadata and correct memory estimates.** Store member metadata once and two indices per active row. Account for grouped functions, extension-field storage, lookup writers, and witness metadata in prover-memory estimates.
- **Consolidate byte tables.** Express byte AND/OR through XOR lookups and byte comparison through subtraction, sharing their existing table multiplicities.
- **Shrink u32 comparison witnesses.** Replace twelve byte auxiliaries with six u16 limbs while retaining six range lookups and bounded inputs. Add the distinct scalar u16 range channel needed by these comparisons.
- **Remove the intermediate u32 comparison carry constraint.** Express the result directly as the whole-word carry `(b + c - a) / 2^32`, with `c = a.wrapping_sub(b)`. This replaces two boolean carry constraints with one, without allocating a carry column. Keep all six limb range checks and both input-recomposition constraints: they bound the equation so field wraparound cannot admit a false comparison.

Cycle defenses and their associated ranks, completion timestamps, component analysis, and unit counters are excluded. No IxVM program optimizations are ported, including the BLAKE3 reader change labeled `perf(aiur)` in #631. Packed query storage is also excluded: its encoding and decoding costs trade execution speed for memory without lowering FFT costs. The generated Rust executors are updated for Aiur's comparison layout; `Ix/IxVM` source is unchanged.

All 83 kernel FFT pins were measured again after rebasing onto Ixon v3 (`519d2c49`) and decrease by **6.6–16.3%** relative to that `main` baseline. These are modeled FFT costs, not wall-clock speedups. The whole-word carry-constraint removal leaves the FFT costs unchanged; it removes one polynomial constraint per comparison, with no runtime improvement measured yet. The measured pin refresh is the final commit.

| Case | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| `Nat.add_comm` | 322,598,714 | 286,989,161 | 11.0% |
| Shard pipeline | 7,072,190,269 | 6,434,780,974 | 9.0% |

Validation passes:

- All 52 active Aiur Rust tests, including exhaustive byte/u16 lookup checks, supplied-trace rejection tests for each comparison limb bound and field wraparound, verifying-key roundtrips, and grouped-witness tests.
- Lean suites: `aiur-cross`, `aiur-cost`, `aiur-prove`, `aiur-hashes`, `rbtree-map`, `multi-stark`, `recursive-verifier`, and `ix-aggr`. Recursive verification explicitly exercises quotient degrees two and four in one proof.
- Full `ixvm` suite: 843 passing assertions, including refreshed pins, adversarial fixtures, and all 249 interpreter/generated-executor parity assertions.
- All three generated executors pass `lake exe ix codegen --check`.
- `cargo fmt --all -- --check` and strict workspace Clippy:

```sh
cargo clippy --locked --release --workspace --all-targets \
  --features ix-ffi/parallel,ix-ffi/net,ix-ffi/test-ffi -- -D warnings
```


arthurpaulino:
!benchmark fresh

arthurpaulino:
!benchmark fresh

arthurpaulino:
!benchmark fresh

johnchandlerburnham (approved):

PR #639, opened 2026-09-17
Recover Ixon v3 compilation and anonymous kernel-check performance

johnchandlerburnham:
The Ixon v3 changes in #636 added repeated declaration preparation and decoding to compilation. The definition-dependency guard in #637 also made anonymous kernel checking traverse each safe definition's complete dependency closure. Whole-environment checks consequently materialized and scanned the same external declarations many times, even when those declarations were unnecessary for checking the current item.

This PR shares compiler preparation work and establishes an explicit checking boundary around decoded Ixon environments. Local definition cycles are rejected during kernel ingress, allowing anonymous checking to preserve lazy external dependencies. In local runs with the default memory admission policy, Mathlib compilation falls from 289.171 seconds to 65.323 seconds, and Init + Std checking falls from 836.832 seconds to 8.459 seconds. A separate pre-v3 comparison puts Mathlib back near its earlier compilation time.

### Reuse the parallel setup decode for semantic validation

Previously, `compile_env_with_profile` decoded every declaration in a serial semantic-contract validation pass. The existing parallel setup scan then decoded every declaration again to build the reference graph, determine immediate groundedness, and collect inductive groups.

Semantic-contract validation now runs inside that setup scan, using the declaration already decoded for the other setup work. The scan returns both the semantic declaration set and any validation error. Invalid semantic contracts still stop compilation before scheduling begins; removing the separate pass does not bypass validation.

### Avoid contract resolution and decoration for ordinary declarations

The Lean frontend now checks both reserved metadata namespaces in one shared-expression traversal. It also checks selected source-contract registrations and measure hints. When none are present, preparation returns the original declaration list directly.

Large inputs are scanned in batches of 1,024 declarations on Lean's worker pool. Each batch only reads immutable declarations and returns a presence flag. Small inputs run directly. This avoids a serial scan of every expression before the parallel Rust compiler starts; complete source-contract preparation now takes about 1.3–1.4 seconds for Mathlib. Verbose compilation reports that phase separately.

The fast path retains declaration-name and uniqueness checks and examines declaration types, bodies, and recursor-rule bodies. A registration or reserved metadata marker selects the full resolution, validation, and decoration path, including malformed-marker rejection.

### Make decoded Ixon ingress the checking boundary

`IxonChecker` accepts an already decoded `ixon::Env` and owns a private anonymous `KEnv`. The data path is:

```text
.ixe bytes → Ixon decoding → verified kernel ingress → typechecking
```

Callers cannot insert arbitrary kernel declarations into this worker. Anonymous FFI checking, profiling, resource validation, and the Zisk and SP1 guest callers now use this interface. Fuel accounting, profiling, diagnostics, and the existing cache-clearing policies remain available through the wrapper.

### Validate local recursion before admitting a block

Ixon represents external dependencies with content addresses and local recursion with `Rec` slots. Ingress now collects `Rec` edges while converting definition types and values. Before publishing a standalone definition or any member of a mutual block, it rejects local cycles reachable from safe definitions.

This includes references encountered through shared expressions. Acyclic forward references remain valid, and the existing treatment of unsafe and partial definitions is preserved. A rejected block publishes no declarations or block entries.

`LazyConstant::get_at` explicitly verifies the binding between a constant's bytes and its environment map key. Anonymous ingress applies it to standalone constants, mutual blocks, and every projection admitted with a block. This also covers environments assembled in memory and prevents an address-verification result from being reused for a different key.

With those bindings established, a cycle between distinct serialized blocks would require solving a cryptographic hash preimage problem. The remaining representable back edges are the local `Rec` edges checked during ingress. `IxonChecker` can therefore skip the repeated traversal of external definition closures.

The general `TypeChecker` API still performs the defensive dependency traversal for raw, mutable `KEnv` callers. Checking an individual Ixon item also retains the existing claim semantics: referenced declarations are assumptions until their own work items are checked. Certifying a complete environment still requires checking every work item.

## Performance

These are local measurements on the same Ryzen 9 7950X3D machine, using release builds, Lean 4.33.1, cached imports, and 32 workers. The following table uses the default adaptive memory admission policy. The baseline is `bcc69954b9b177752e70611d895b81046aa49db1`, which already includes both regressions. Compile times cover declaration preparation, Rust compilation, and file writing; initial import loading is outside the CLI's compile timer.

| Workload | Baseline | This PR | Speedup |
| --- | ---: | ---: | ---: |
| Compile Init + Std: 117,084 source constants | 25.358 s | 10.317 s | 2.46× |
| Compile Mathlib: 771,135 source constants | 289.171 s | 65.323 s | 4.43× |
| Check Init + Std: 98,795 kernel constants | 836.832 s | 8.459 s | 98.9× |

Init + Std checking peak RSS falls from 6.355 GB to 1.794 GB.

Compiled outputs are byte-for-byte identical to the baseline: 342,786,227 bytes for Init + Std and 3,343,274,312 bytes for Mathlib. All 672,981 Mathlib kernel constants also pass checking.

A separate comparison against the pre-v3 revision, `cf77c957e50d64a5ed42330d3ae8c176296e7d1d`, checks how much of the original performance is recovered:

| Mathlib compile configuration | Pre-v3 | This PR |
| --- | ---: | ---: |
| Default adaptive admission | 64.368 s | 65.323 s |
| Fixed 32 workers (`IX_COMPILE_ADAPTIVE=0`) | 54.523 s | 53.856–57.880 s |

The fixed-worker range contains two runs. These are individual local measurements, not a statistical benchmark campaign. Fixed admission was used to isolate compiler work from scheduler pauses; the PR retains the default admission policy. The optimized v3 outputs match the v3 baseline byte for byte, while the pre-v3 output remains its original v2 encoding.

A separate Mathlib kernel comparison used clean release binaries, each version's matching `.ixe`, and 32 workers, with no concurrent builds or benchmarks:

| Mathlib kernel configuration | Pre-v3 | This PR | Difference |
| --- | ---: | ---: | ---: |
| Normal whole-environment check | 118.866 s | 119.674 s | +0.7% |
| Per-constant profiling enabled | 139.325 s | 146.074 s | +4.8% |

All 672,981 constants passed in every run. Normal checks were run in reverse order relative to the profiled pair. The profiles contain 663,254 work items each; aggregate heartbeats, substitution nodes, reduction calls, equality calls, arithmetic work, and intern visits differ by less than 0.001%. These measurements do not reproduce the historical large Mathlib checking slowdown, though the small timing differences are insufficient to establish or exclude a minor regression.

## Validation

- **Rust:** 1,501 compiler, codec, and kernel tests passed; 14 existing tests remain ignored.
- **Static checks:** Full workspace Clippy passed with `parallel`, `net`, and `test-ffi` enabled. Formatting and whitespace checks passed.
- **Lean:** The build passed with warnings treated as failures, and the full primary Lean test suite passed. Source-contract tests and the focused definition-dependency suite passed, including its 44 checks for malformed cyclic inputs, acyclic forward references, and compiled structural, well-founded, and mutual recursion.
- **Ixon v3:** Integration tests passed, including codec parity, rejection cases, VM proofs, and primitive-closure validation. Generated-code checks passed.
- **Regression coverage:** New tests verify semantic validation in the shared setup scan, the ordinary-declaration preparation path, metadata detection at parallel batch boundaries and in the final partial batch, lazy checking of a 1,025-definition chain, atomic rejection of local cycles, acceptance of an acyclic mutual block, and rejection of invalid constant or sibling-projection address bindings.
- **Guest integration:** SP1 passed host type checking. Zisk passed a host compatibility check using temporary input-buffer borrows required by its host API; those temporary changes and both guest lockfiles were restored. Target-specific Zisk and SP1 SDKs were unavailable, so native guest execution and guest cycle measurements were not performed.

To reproduce the primary measurements in each checkout, prepare the cached import artifacts and retain separate output files:

```sh
lake build ix
.lake/build/bin/ix compile Benchmarks/Compile/CompileInitStd.lean \
  --no-build --out InitStd.ixe --json compile-initstd.json
.lake/build/bin/ix compile Benchmarks/Compile/CompileMathlib.lean \
  --no-build --out Mathlib.ixe --json compile-mathlib.json
.lake/build/bin/ix check-rs InitStd.ixe --anon --workers 32 \
  --json check-initstd.json --json-name InitStd
.lake/build/bin/ix check-rs Mathlib.ixe --anon --workers 32 \
  --json check-mathlib.json --json-name Mathlib
```

Compare each pair of compiled environments with `cmp` to verify byte equality.


samuelburnham:
!benchmark fresh

samuelburnham:
!benchmark compile fresh

samuelburnham:
!benchmark ooc fresh

samuelburnham (approved):

PR #640, opened 2026-09-18
Improve Aiur compilation and native execution

arthurpaulino:
Reduce redundant compiler-generated work and native query-table overhead,
then separate Rust syntax construction from compact source emission. This
changes the Aiur compiler and execution infrastructure, not the IxVM's
typechecking or reduction algorithms.

## Changes

- **Evaluate match scrutinees exactly once.** Bind non-variable scrutinees
  before pattern compilation. Named fallbacks previously could evaluate them
  again; wildcard-only matches could skip them, including assertions and IO.
- **Reuse pure values within bytecode blocks.** Reuse identical constants,
  field arithmetic, zero tests, and immutable memory operations; forward loads
  from retained stores. Availability is confined to each block and restored
  across branches. Calls, assertions, IO, range checks, debug operations, and
  unconstrained hints are excluded from this reuse pass. The first memory
  operation remains, preserving allocation order and invalid-load rejection.
- **Use fixed native memory-table slots.** Replace repeated width-keyed
  lookups with indexed access, validating the table layout once at entry,
  before any mutation. Pointer checks and row ordering remain intact.
- **Reuse query-key hashes.** Carry the hash from lookup into function return
  registration and memory insertion. Full-key equality is still checked, and
  return registration searches again after the body: nested execution may
  have inserted the same key or grown the table. Constrained calls still
  replay cached hint bodies when promotion is required.
- **Specialize native execution modes.** Use a const-generic constrained /
  unconstrained flag so Rust can eliminate mode-dependent branches. Preserve
  stack-growth guards and query accounting.
- **Build Rust through structured syntax.** Explicit nodes represent types,
  patterns, blocks, guarded matches, closures, methods, casts, error
  propagation, unsafe blocks, and functions. Remove executable string
  fragments and source-text import scanning; centralize precedence and string
  escaping. Output remains compact: one line per top-level item.

Regenerate the IxVM, recursive-verifier, and aggregator executors. The large
generated-file deletion count is primarily compaction, not removed execution
logic. Update all 83 constant FFT pins and the shard-pipeline pin for the
compiler improvements; every pin decreases. Native-only changes leave them
unchanged.

## Measurements

For native `ix check Lean.Json`, compiler changes reduce FFT cost from
**190,226,521,931 to 189,750,808,287 (-0.25%)**. The local-reuse pass reduces
total circuit width from **12,439 to 12,419**, including
`blake3_compress_block` from **201 to 191 columns**. Compiler-only timings do
not establish an execution-speed improvement.

Native optimizations were measured incrementally against their immediate
predecessors, using sequential paired runs with `RAYON_NUM_THREADS=32` and
`/usr/bin/time -v`, without concurrent builds or tests:

| Change | Paired runs | Mean before | Mean after | Reduction |
|---|---:|---:|---:|---:|
| Fixed memory-table slots | 3 | 15.183 s | 14.717 s | 3.1% |
| Query-key hash reuse | 3 | 15.037 s | 14.263 s | 5.1% |
| Execution-mode specialization | 6 | 14.287 s | 14.173 s | 0.8% |

These are whole-command wall times, not isolated executor timings. The mode
specialization result is marginal, with overlapping timings. The percentages
come from separate batches and should not be added into an overall speedup.
Peak RSS was effectively unchanged. Complete circuit-statistics reports were
identical within each native-only comparison and across the generator refactor.

## Validation

- Rebuilt `ix` and `IxTests`; `ix codegen --check` confirms all three generated
  executors are current.
- Full ignored IxVM suite passed: **249 native/interpreter parity assertions
  and 84 FFT pins**, with the final full run completing in **3:34.95**.
- Compiler cross-evaluator and prove/verify tests, MultiStark primitive tests,
  recursive verification, and aggregation tests passed.
- Added four behavioral query-map tests covering nested insertion/table
  growth, same-key insertion after a miss, empty keys/outputs, promotion, and
  equivalence with ordinary registration. All pass in release mode.
- Added a quick Rust syntax test that compiles and executes generated code,
  rather than comparing source snapshots. It covers precedence, escaping,
  guard fallthrough, both error-propagation outcomes, const generics, unsafe
  array access, and generated debug output.
- Repaired the recursive-verifier coverage fixture: put loads behind a call
  boundary so local forwarding cannot erase the lookups intended to exercise
  quotient degree four. Its degree-two/degree-four coverage assertion, honest
  proof verification, native parity, and tampering checks now all pass.
- Release Clippy for `ixvm-codegen` passes with `-D warnings`. Compact output
  matches the validated multiline output exactly outside whitespace, with
  string contents preserved; compact native build and syntax checks pass.


arthurpaulino:
!benchmark fresh

samuelburnham (approved):

arthurpaulino:
!benchmark fresh

arthurpaulino:
!benchmark fresh

